PR #1, opened 2018-02-14
update links regarding version info and changelog

indirectlylit:
* remove the warning that kolibri is not released
 * link to changelog
 * remove broken `include`
 * link to other docs versions

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -3,10 +3,7 @@
 Install Kolibri
 ###############
 
-
-.. warning::
-  **Kolibri** is not released yet, so these guides are *Work in Progress* !
-
+These docs are for version 0.7.x. See the `Kolibri changelog here <https://learningequality.org/r/kolibri-changes>`_. Documentation for `other versions are available here <https://readthedocs.org/projects/kolibri/>`_.
 
 To install **Kolibri**, check the system requirements first and then follow the procedure for the operating system on your device.
 
@@ -31,5 +28,3 @@ To install **Kolibri**, check the system requirements first and then follow the 
 
 .. include:: install/_initial_setup.rst
 
-
-.. include:: ../CHANGELOG.rst


benjaoming:
Sorry I hadn't seen that the inclusion of the change log had moved to this location. Should we include it still?

I hard-copied it into the previous versions, knowing where it was linked.

docs/install.rst
```diff
@@ -31,5 +28,3 @@ To install **Kolibri**, check the system requirements first and then follow the
 
 .. include:: install/_initial_setup.rst
 
-
-.. include:: ../CHANGELOG.rst
```
benjaoming, line 35: Sorry I hadn't seen that the inclusion of the change log had moved to this location. Should we include it still?

I hard-copied it into the previous versions, knowing where it was linked.

indirectlylit: In the [link above](https://github.com/learningequality/kolibri-docs/pull/1/files#diff-62c330cd8da31d476bb63b51bb8b1307R6), I created a new redirect link to `https://learningequality.org/r/kolibri-changes`.

My thinking is that this will help us maintain a single source of truth, and also be able to change the location after-the-fact.

thoughts?

benjaoming: That makes sense regarding single-source-of-truth!

One thought/objection is about a disadvantage - the documentation is versioned and the link isn't (and currently links to Github). So the user will always have to leave the documentation, either by means of going to a different platform (the Github repo / leq homepage) or stay in the documentation, but being switched to another documentation version (which is also bad)

I would suggest to either:

1) Keep maintaining user release notes in the user docs -- the authoring of these can come in hind-sight, as with the rest of the documentation. I can add something in the release process about creating a new user-docs branch and to compile and edit release notes here.

2) Use links like `https://learningequality.org/r/kolibri-changes-0.7.x` -- but this would then likely link to a specific branch of `kolibri-docs` or `kolibri`, which doesn't solve much.

benjaoming: Oh maybe if you point `https://learningequality.org/r/kolibri-changes` to `https://kolibri.readthedocs.io/latest/changes.rst`, we could keep the single-source-of-truth + ability to restructure docs (which we seem to still enjoy!) ?

indirectlylit: My assumption was that linking to the most recent changelog would be fine, even for older versions of the docs?

benjaoming: Alright, let's do that -- but I'd really like if we aimed at having a *user* changelog and a *developer* changelog :)

benjaoming:
Please don't delete your branch, just gonna fetch your changes, make them work for current `developer` and get them in...

benjaoming:
Changes are now in #4

PR #12, opened 2018-04-04
update linux support to indicate python 3.4

indirectlylit:
re: https://github.com/learningequality/kolibri-installer-debian/issues/25

currently we indicate that python 2.7 is supported, but it's not

### DIFF
--- a/docs/install/_system_requirements.rst
+++ b/docs/install/_system_requirements.rst
@@ -8,7 +8,7 @@ Operating systems
 
  - Windows Vista, 7, 8.1, 10
  - Debian/Ubuntu: 14.04, 16.04 and up - anything that's *not* end-of-life.
- - Linux: Any system with Python 2.7
+ - Linux: Any system with Python 3.4
  - Mac OSX 10.9, 10.10 and 10.11
  - Debian/Raspberry Pi packages: Wheezy or later
  - *(planned)* Android: 4.2.2 or later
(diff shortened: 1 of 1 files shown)


docs/install/_system_requirements.rst
```diff
@@ -8,7 +8,7 @@ Operating systems
 
  - Windows Vista, 7, 8.1, 10
  - Debian/Ubuntu: 14.04, 16.04 and up - anything that's *not* end-of-life.
- - Linux: Any system with Python 2.7
+ - Linux: Any system with Python 3.4
```
benjaoming, line 11: It should say "Any system with Python 2.7 and 3.4+". The issue is that the Debian installer is only built for Python3, which in any case is available on the systems listed above.

benjaoming (changes requested):
Python 2 and 3 are both supported on linux. It's a matter of which the Debian is built for, e.g. it installs to `/usr/lib/python3.5/dist-packages`

indirectlylit:
Okay, so is it this line that needs to be updated, e.g.:

> - Debian/Ubuntu with Python 3.4: 14.04, 16.04 and up - anything that's *not* end-of-life.

benjaoming:
@indirectlylit yes, that would be more accurate. I think the problem with Tim's installation is that he doesn't want Python 3.5 on the system and has removed it completely, so with `dpkg -i` the missing dependency isn't installed but instead causes a halt (because `dpkg` does not install dependencies).

indirectlylit:
closing. after looking at the docs, I think we need a bit more work than just this change. new issue coming.

Issue #15, opened 2018-04-07
installation instructions and requirements are confusing/misleading


indirectlylit:
There are a few things that are confusing:

1. PEX is framed as just a mac installation method. However it also works on linux.
2. PIP is framed as an 'advanced' installation method and contains the phrase "Once Kolibri is released". There's no indication that it works on Mac, Windows, and Linux
3. The linux system requirements are very confusing
   * There are three independent-looking items, Ubuntu, Debian/Raspbian, and Linux. However these are not mutually exclusive categores (Linux > Debian > Ubuntu). This is leading to [confusion like this](https://github.com/learningequality/kolibri-installer-debian/issues/25)
   * The Debian support is much more limited than PEX and PIP support, and this isn't indicated. Sounds like it's currently only Python 3.4 and 3.5? ([see also this](https://github.com/learningequality/kolibri-installer-debian/issues/29))

Seems like we might need to better distinguish installation _methods_ (pip, .deb, pex, etc) from install targets (debian linux variants, osx, general linux, etc).

cc @radinamatic  @benjaoming

radinamatic:
Good points @indirectlylit, I'll think about best way to rework the system requirements and installation pages for Linux and Mac.

benjaoming:
Fixed in #16

PR #14, opened 2018-04-07
Add very basic instructions to get Kolibri running on Raspberry Pi

benjaoming:
@radinamatic this is really urgent IMO because Kolibri currently does **not** work on Raspbian without these special steps.

Please feel free to make any changes to this!

@lyw07 FYI :)

PR #16, opened 2018-04-11
Improve the system reqs and installation instructions

radinamatic:
Addresses #15.

- [X] PEX is framed as Mac & other Linux distros installation method
- [X] PIP is still framed as an 'advanced' installation method with indication that it works on Mac, Windows, and Linux
- [X] Linux system requirements are still complex, but hopefully more accurate now
- [ ] The Debian support is *still* currently [only Python 3.4 and 3.5](https://github.com/learningequality/kolibri-installer-debian/issues/29). Do I add warning for it, or are we going to include support Python > 3.6? cc @aronasorman 

I Included @benjaoming's instructions for Raspberry from #14 so it can be merged together.


docs/install/_advanced.rst
```diff
@@ -8,44 +8,21 @@ Advanced installation options
...
 .. _raspberry-pi-wifi:
 
 Raspberry Pi
 ------------
 
 .. note::
-  This section is pending instructions for creating a WI-FI hotspot. You can
-  install Kolibri on Raspbian following the above instructions for the PPA or
-  the ``pip install`` method.
+  This section is pending instructions for creating a WI-FI hotspot on Raspberry Pi. Follow the instructions above to install Kolibri on :ref:`Raspbian <rpi>`.
```
indirectlylit: There two Raspberry Pi sections (here and under Linux). Should probably just delete this one?

The fact that we eventually want to have instructions on setting up a wifi hotspot is not helpful

docs/install/_advanced.rst
```diff
@@ -8,44 +8,21 @@ Advanced installation options
...
 -------------------
 
 .. note::
-  This section is pending instructions for NGINX configuration. You can
-  install Kolibri on Raspbian following the above instructions for the PPA or
-  the ``pip install`` method. If you are an experienced system administrator or
-  Django user, know that it is possible to use Kolibri with for instance
-  UWSGI+Nginx, using conventional methods.
-
+  This section is pending instructions for NGINX configuration. You can install Kolibri on Raspbian following the above instructions for the PPA or the ``pip install`` method. If you are an experienced system administrator or Django user, know that it is possible to use Kolibri with for instance UWSGI+Nginx, using conventional methods.
```
indirectlylit: A few things still bother me about this section...

* it's not clear to me what Raspbian has to do with Ngninx?
* We should standardize on one capitalization, probably 'Nginx'
* having stuff like "This section is pending" to seems unhelpful and just adds noise

-----

image

I'd recommend removing the entire 'advanced' section, and replacing it all with a single top-level "pip install" section.

docs/install/_linux.rst
```diff
@@ -37,18 +39,81 @@ The advantages of downloading a ``.deb`` file is the portability: you can copy t
...
+      sudo pip3 install cffi --upgrade
+      sudo systemctl start kolibri
+
+3. When the command finishes, open the default browser at http://127.0.0.1:8080 and proceed with the :ref:`setup_initial` of your facility. 
+
+.. warning: Loading channels can take a **long time** on a Raspberry Pi. When generating channel contens for Khan Academy, * Generating channel listing. This could take a few minutes…* means ~30 minutes. The device's computation power is the bottleneck. You might get logged out while waiting, but this is harmless and the process will continue. Sit tight!
+
+
+Troubleshooting
+^^^^^^^^^^^^^^^
```
indirectlylit: Recommend making this an 'info' block so that it doesn't add noise to the TOC:

image

indirectlylit:
> The Debian support is still currently only Python 3.4 and 3.5. Do I add warning for it, or are we going to include support Python > 3.6?

I think we should add a warning until we change what's supported. It's easy to update the docs so better to just make the accurate to the current state of the system?

As far as information structure for 'system requirements', here's my recommendation:

1. Move all client-side requirements to a different page, because they're not really  relevant here
   * Maybe more appropriate in the 'Access Kolibri' page, near the 'other devices' section?
   * Let's consolidate the 'Clients' and 'Supported Browsers' sections - as far as I can tell, they are the same thing

2. Instead of having a list of supported server operating systems at the top, maybe it would be easier to scatter the requirements across the various install methods? Then people first just look for their OS, and then see only the relevant instructions and compatibility?


Here's an example of what the final TOC might look like:


**Install Kolibri**

* Windows
  * Compatibility
  * Install
  * Uninstall
  * Upgrade
* Debian/Ubuntu
  * Compatibility
  * Install from PPA
  * Install from .deb file
  * Uninstall
  * Upgrade
* Raspbian
  * Compatibility
  * Install
  * Uninstall
  * Upgrade
* Other Linux & Mac
  * Compatibility
  * Install
  * Uninstall
  * Upgrade
* Python pip package
  * Compatibility
  * Install
  * Uninstall
  * Upgrade

indirectlylit:
Also, maybe we should just remove the Android instructions, and consolidate Mac with 'Other Linux'?

benjaoming:
As soon as this is released, we don't have to mention that .deb doesn't support Python 3.6: https://github.com/learningequality/kolibri-installer-debian/issues/29

If there are problems with Python 3.6, we can treat them as bugs and get them fixed :)

docs/install/_linux.rst
benjaoming: There's an intention to create more sections from my side, possibly turning this into a "How-to" instead.

docs/install/_linux.rst
```diff
@@ -37,18 +39,81 @@ The advantages of downloading a ``.deb`` file is the portability: you can copy t
...
+.. _rpi:
+
+Raspberry Pi (Raspbian)
+-----------------------
+
+Kolibri currently doesn't work out of the box for Raspbian Jessie. We are refining our distribution to work out of the box, but you need to follow these two steps in order to install it (tested on Kolibri 0.9):
+
+1. Running ``add-apt-repository`` as shown in the PPA instructions does not work. Instead, run::
+
+      echo "deb http://ppa.launchpad.net/learningequality/kolibri/ubuntu xenial main" > /etc/apt/sources.list.d/learningequality-ubuntu-kolibri-xenial.list
```
lyw07: @benjaoming I'm currently on a raspberry pi 3 with Raspbian Jessie. I run the command `echo "deb http://ppa.launchpad.net/learningequality/kolibri/ubuntu xenial main" > /etc/apt/sources.list.d/learningequality-ubuntu-kolibri-xenial.list` but it errors out with `bash: /etc/apt/sources.list.d/learningequality-ubuntu-kolibri-xenial.list: Permission denied`. I also try to run the command with `sudo`, but it doesn't work. I think the problem here is that the file `/etc/apt/sources.list.d/learningequality-ubuntu-kolibri-xenial.list` has never been created. So I'm wondering before running this command, it would be better to create a file first? Do you have any suggestions? Thanks!

docs/install/_linux.rst
indirectlylit: Might be best to break that into a separate file then so the top-level page stays simpler

docs/install/_linux.rst
benjaoming: Hi @lyw07 :)

Redirecting output to a file with `>` will create it when it doesn't exist. It's because `sudo` only binds to the first command, `sudo cmd > file` is the same as `(sudo cmd) > file`, `>` parses with a very low priority.

Anyways, this should work: `sudo su -c 'echo "deb http://ppa.launchpad.net/learningequality/kolibri/ubuntu xenial main" > /etc/apt/sources.list.d/learningequality-ubuntu-kolibri-xenial.list'`

lyw07: ah thank you so much! that totally makes sense!

lyw07: @benjaoming I'm also wondering that instead of naming it a `raspberry pi` issue, would it be okay to name it a `debian jessie` issue? I tested on debian jessie days ago and found the same issue there: https://github.com/learningequality/kolibri-installer-debian/issues/30#issuecomment-381781511. My guess is that it's more likely a debian jessie issue?

benjaoming:
Hi Lingyi!

Yes it's definitely a general Debian issue since the PPA doesn't contain
the various Debian dist names, so we have to manually specify "xenial"
or "trusty" in the apt source. Using `apt-add-repository` will
automatically use the current dist name, which isn't sufficient when it
isn't Ubuntu.

Best,
Ben

lyw07:
@benjaoming Thank you for the explanation! I agree that `apt-add-repository` needs to be replaced by manually specifying dist names. But if I remember correctly, `apt-add-repository` works on Debian Stretch as it gets `Ubuntu Bionic`. Have you had problem running `apt-add-repository` on Debian Stretch as well? Thank you!

indirectlylit:
awesome, thanks! what do you think of this organization, @radinamatic?

Btw, travis is [failing](https://travis-ci.org/learningequality/kolibri-docs/builds/370335859?utm_source=github_status&utm_medium=notification):

```
Warning, treated as error:
/home/travis/build/learningequality/kolibri-docs/docs/access.rst:78:Title level inconsistent:
```

radinamatic:
@indirectlylit I tried to include all your suggestions:
- OS/browser reqs are now indicated per each platform, client ones moved to the `Access page`
- Eliminated "Advanced installation" and added `pip install` as top level item
- Added the new `How to...` section, invisible for now, for when we have the RPi-wifi and Nginx sections ready

@lyw07 Can I get your eyes on the `Raspbian` section? Haven't done these steps myself, and your review will be the final word! :slightly_smiling_face:

docs/install/_rpi.rst
```diff
@@ -0,0 +1,55 @@
...
+1. Running ``add-apt-repository`` as shown in the PPA instructions does not work. Instead, run::
+
+      sudo su -c 'echo "deb http://ppa.launchpad.net/learningequality/kolibri/ubuntu xenial main" > /etc/apt/sources.list.d/learningequality-ubuntu-kolibri-xenial.list'
+      sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys DC5BAA93F9E4AE4F0411F97C74F88ADB3194DD81
+      sudo apt update
+      sudo apt install kolibri
+
+2. Kolibri does not start after installation. This is because ``python3-cffi`` is outdated on Raspbian. Upgrade it like this::
+
+      sudo apt install libffi-dev
```
lyw07, line 23: Not sure why we need `libffi-dev`... I tried the installation steps without this step, and kolibri works fine.

radinamatic: That was verbatim the original RPi installation instructions by @benjaoming from #14. If you are positive it's not necessary, I can delete that line.

lyw07: I see. Thanks! Then maybe we can leave it here for now before Ben comes back to see if kolibri works on his machine without this step? The result could differ by different machines.

radinamatic: Sounds good! :+1: 

@indirectlylit Ready for the final review! :wink:

indirectlylit:
awesome, thanks @radinamatic!

docs/access.rst
```diff
@@ -7,7 +7,7 @@ Access Kolibri
 Starting Kolibri on Windows
 ===========================
 
-To start **Kolibri** on Windows, just double-click the desktop shortcut. You will see the notification message *Kolibri is starting, please wait...*.
+To start **Kolibri** server on Windows, just double-click the desktop shortcut. You will see the notification message *Kolibri is starting, please wait...*.
```
indirectlylit: > To start Kolibri server on Windows

Maybe "To start **the** Kolibri server on Windows"?

docs/access.rst
```diff
@@ -53,40 +61,49 @@ Starting Kolibri on Linux
...
+*************
 
-  Tap the Kolibri icon among your device apps to start.
+Kolibri currently supports these combinations of operating systems and browsers as client devices. Videos are MP4 encoded, and require no additional installation of plugins or codecs.
 
+* Windows 7, 8.1 and 10, with IE 11+, Chrome and Firefox
+* MacOS 10.6+ with Safari, Chrome and Firefox
+* Linux, any browser
+* Android 4.2+, Chrome and Firefox
+* iOS ``XXX``, Chrome and Firefox supported, **Safari not supported**
```
indirectlylit: maybe remove ``XXX``

docs/install.rst
```diff
@@ -3,9 +3,9 @@
 Install Kolibri
 ###############
 
-See the `Kolibri changelog here <https://learningequality.org/r/kolibri-changes>`_. Documentation for `other versions are available here <https://readthedocs.org/projects/kolibri/>`_.
+See the `Kolibri changelog here <https://learningequality.org/r/kolibri-changes>`_. Documentation for `previous Kolibri versions are available here <https://readthedocs.org/projects/kolibri/>`_.
 
-To install **Kolibri**, check the system requirements first and then follow the procedure for the operating system on your device.
+To install **Kolibri** on device that will act as server, check the hardware requirements first and then follow the procedure for the operating system on your device.
```
indirectlylit: >  on device that will act as server

maybe "on **a** device that will act as **a** server"?

docs/install/_rpi.rst
```diff
@@ -0,0 +1,55 @@
+.. _rpi:
+
+Raspbian
+========
+
+Compatibility
+-------------
+
+Kolibri currently doesn't work out of the box for Raspbian Jessie. We are refining our distribution to work out of the box, but you need to follow the below steps in order to install it (tested on Kolibri 0.9).
```
indirectlylit: > but you need to follow

"but you can follow"

docs/install/_windows.rst
```diff
@@ -3,7 +3,12 @@
 Windows
 =======
 
-To install or uninstall Kolibri on Windows, follow these steps.
+Compatibility
+-------------
+
+* Supported: Windows 7, 8.1 and 10, with IE 11+, Chrome or Firefox
+* **Not supported: Windows XP** cannot be used to install Kolibri server, but could potentially work as a client device if the browsers are `as up-to-date as possible <https://support.mozilla.org/en-US/questions/1173904>`_.
```
indirectlylit: The bolding on

```
**Not supported: Windows XP** cannot be used
```

is a bit strange. Maybe

```
**Not supported:** Windows XP cannot be used
```

docs/manage/_how_to.rst
```diff
@@ -0,0 +1,20 @@
...
+Stay tuned!
+
+
+.. _nginx:
+
+Configure Nginx
+---------------
+
+.. note::
+  If you are an experienced system administrator or Django user, know that it is possible to use Kolibri with for instance UWSGI+Nginx, using conventional methods.
```
indirectlylit: > If you are an experienced system administrator or Django user, know that it is possible to use Kolibri with for instance UWSGI+Nginx, using conventional methods.

Maybe

> If you are an experienced system administrator or Django user, it is possible to deploy Kolibri using conventional methods - for instance with uWSGI and Nginx.

indirectlylit (approved):
Looks great! just a few minor tweaks

docs/access.rst
radinamatic: No version number on iOS?

radinamatic:
@indirectlylit Done, done & done! :slightly_smiling_face:

benjaoming:
Hey, I need the RPi instructions to be live, but can see that the final review from @indirectlylit  is marked done - hoping that it's okay to merge :)

docs/access.rst
indirectlylit: I don't think we know...

indirectlylit:
great upgrade - thanks all

PR #18, opened 2018-04-25
Add Material Design icons to user docs

radinamatic:
Makes it easier to use the same set of icons in the docs as those in Kolibri itself.

Some "massaging" of the padding was necessary to get the acceptable alignment.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -110,7 +110,7 @@ html_logo = 'logo.png'
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
-
+html_style = 'theme_overrides.css'
 # This should be commented back in for wide tables
 # See: https://github.com/rtfd/readthedocs.org/issues/2116
 # and: https://github.com/rtfd/sphinx_rtd_theme/pull/432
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -125,9 +125,9 @@
 # adapted from: http://rackerlabs.github.io/docs-rackspace/tools/rtd-tables.html
 # and https://github.com/altair-viz/altair/pull/418/files
 # https://github.com/rtfd/sphinx_rtd_theme/issues/117
-def setup(app):
+# def setup(app):
     # Add our custom CSS overrides
-    app.add_stylesheet('theme_overrides.css')
+  #  app.add_stylesheet('theme_overrides.css')
```
indirectlylit, line 130: Is the setup function no longer necessary?

If so, we should delete it and the comment above. However, it's not clear to me why the change was made?

indirectlylit (commented):
this seems to work for me, but I'm not entirely sure because one of the two configuration files still references the older `fa` code. (see #19)

If we're transitioning to Material, Let's also make sure we're deleting the old fontawesome files so that the browser doesn't need to download them unnecessarily.

PR #37, opened 2018-08-22
Update .po files, and makefile

indirectlylit:
* cleaned up old .po files
* updated spanish to current translated state
* cleaned up inconsistencies in makefile


benjaoming (approved):
LGTM! Thanks @indirectlylit !

PR #43, opened 2018-08-27
Capture that develop is 0.11 and in develop mode

benjaoming:
@radinamatic I don't know if this is any good, but let's work on `develop` for 0.11 so the branch makes sense to have.


radinamatic:
Roger that @benjaoming! 👍

PR #47, opened 2018-09-10
Move `es` locale images to `docs/img/es` and add `figure_language_filename`

### LINKED ISSUE #46: Translated figures (screenshots) not showing
I think we need to consolidate the configuration with the documentation for `figure_language_filename`:

http://www.sphinx-doc.org/en/master/usage/configuration.html#confval-figure_language_filename

benjaoming:
Tested locally, but not yet on RTD - will merge and see if has the same use of `es` (and not `es_ES`).

Fixes #46


radinamatic:
If I understand correctly, this change assumes only the path change, not the filename itself?

benjaoming:
@radinamatic not sure if I understand the question?

radinamatic:
Just trying to understand better the change `figure_language_filename = '{path}{language}/{basename}{ext}'`. If according the [specs](http://www.sphinx-doc.org/en/master/usage/configuration.html#confval-figure_language_filename), the default (which we had so far):

> {root}.{language}{ext} will be expanded to dirname/filename.**en**.png from `.. image:: dirname/filename.png`.

In our case it expands to the same value `dirname/filename.png` in both English and Spanish locales:

### EN


### ES


I'm only wondering if we need to change the filenames of the localized images too, like `dirname/filename.ES.png`?

benjaoming:
@radinamatic - no I maintained the old structure so we wouldn't have to go along with the defaults and won't have to change any file names :)

benjaoming:
Btw it seems that there are some files left, which aren't used anymore. Or, I found one by coincidence: `import-local-drive.png`

benjaoming:
@radinamatic also, I get your point: Why would it not have worked before, when the defaults clearly are different from both the `es/` and `es_ES` location of the images.. I have no idea!

radinamatic:
It always makes me perversely happy when _my_ confusion results in being _somebody else's_ confusion too... 😎 

> it seems that there are some files left, which aren't used anymore. 

Ah, yes please, delete all you are sure that are not used anymore... I'm still looking for a better way to detect those before committing... 😒

benjaoming:
I don't think they can easily be spotted retroactively.. probably best to do when removing the `figure`.

benjaoming:
@radinamatic the Spanish screenshots are back!

radinamatic:
Oh, beautiful sight!!!! 😍 
Thank you! 🙇‍♀️

PR #50, opened 2018-09-16
Add chapter about troubleshooting with a firewall

benjaoming:
Based on recent support on '#kalite' IRC, I want to add this for Kolibri as well.

### DIFF
--- a/docs/manage/_get_support.rst
+++ b/docs/manage/_get_support.rst
@@ -60,6 +60,16 @@ About IP addresses
 * ``http://192.168.1.1:8080`` means: "Connect to IP address ``192.168.1.1`` on port ``8080`` with the HTTP protocol". The browser will the continue to try to reach this address, but may fail for instance if Kolibri isn't running, or if a step along the way blocks access.
 
 
+.. _firewalls:
+
+Firewalls (Windows)
+*******************
+
+Especially Windows-based systems often come with firewalls bundled and enabled. This is critical to the server running Kolibri, as it can block traffic in and out of Kolibri. You may want to have your firewall enabled for various security issues, especially if the server is also online.
+
+In order to troubleshoot network issues, you should temporarily disable your firewall to narrow down the problem.
+
+
 Troubleshoot Database Issues
 ----------------------------
 
--- a/docs/manage/_get_support.rst
+++ b/docs/manage/_get_support.rst
@@ -62,12 +62,14 @@ About IP addresses
 
 .. _firewalls:
 
-Firewalls (Windows)
-*******************
+Firewalls
+*********
 
-Especially Windows-based systems often come with firewalls bundled and enabled. This is critical to the server running Kolibri, as it can block traffic in and out of Kolibri. You may want to have your firewall enabled for various security issues, especially if the server is also online.
+If you are having trouble connecting to Kolibri from other computers, your firewall might be blocking access.
 
-In order to troubleshoot network issues, you should temporarily disable your firewall to narrow down the problem.
+Windows systems often come with firewalls bundled and enabled, and this may interfere with running Kolibri. That said, you probably want to keep your firewall enabled for security reasons, especially if the server is connected to the public internet.
+
+You can try temporarily disable your firewall to see if it helps with connecting to Kolibri. If so, you'll want to turn the firewall back on and then create a exception rule for Kolibri to allow access.
 
 
 Troubleshoot Database Issues


docs/manage/_get_support.rst
```diff
@@ -60,6 +60,16 @@ About IP addresses
...
 
 
+.. _firewalls:
+
+Firewalls (Windows)
+*******************
+
+Especially Windows-based systems often come with firewalls bundled and enabled. This is critical to the server running Kolibri, as it can block traffic in and out of Kolibri. You may want to have your firewall enabled for various security issues, especially if the server is also online.
+
+In order to troubleshoot network issues, you should temporarily disable your firewall to narrow down the problem.
```
indirectlylit: recommend updating the wording a bit:

> If you are having trouble connecting to Kolibri from other computers, your firewall might be blocking access.
> 
> Windows systems often come with firewalls bundled and enabled, and this may interfere with running Kolibri. That said, you probably want to keep your firewall enabled for security reasons, especially if the server is connected to the public internet.
> 
> You can try temporarily disable your firewall to see if it helps with connecting to Kolibri. If so, you'll want to turn the firewall back on and then create a exception rule for Kolibri to allow access.

indirectlylit (approved):
thanks for adding this.

My only concern is that the [support and troubleshooting](https://kolibri.readthedocs.io/en/latest/manage.html#support-and-troubleshooting) section is becoming a grab-bag of random issues inside the 'Manage Kolibri' page.

I'd recommend:

1. Move the [Report a problem with Kolibri](https://kolibri.readthedocs.io/en/latest/manage.html#report-a-problem-with-kolibri) information to the existing (and more visible) [Get in touch](https://kolibri.readthedocs.io/en/latest/contact.html) page
2. Move the lengthy [About IP addresses](https://kolibri.readthedocs.io/en/latest/manage.html#about-ip-addresses) section to its own sub-page, or cut it way down in length
3. Rename the section to just 'Troubleshooting'

docs/manage/_get_support.rst
```diff
@@ -60,6 +60,18 @@ About IP addresses
 * ``http://192.168.1.1:8080`` means: "Connect to IP address ``192.168.1.1`` on port ``8080`` with the HTTP protocol". The browser will the continue to try to reach this address, but may fail for instance if Kolibri isn't running, or if a step along the way blocks access.
 
 
+.. _firewalls:
+
+Firewalls (Windows)
```
radinamatic: Nitpicking: isn't there just **a** (one) Windows firewall built in, and enabled by default? Why plural (here and below)?

radinamatic: Mostly because in the last paragraph you mention disabling **just one**, so implying there is more than one might confuse the user (_"How do I (temporarily) disable the others?"_).

indirectlylit: yeah .. I think the confusion is whether this applies only to Windows or not

Is this section about "the firewall on windows" or is it about "firewalls which may or may not be installed on your operating system"

Maybe we should make it the latter?

benjaoming: I just wanna catch the attention of especially Windows users with the headline. I know that other systems may have firewalls as well, but they have to be installed somewhat knowingly.

For Windows, the Firewall is often the one that is built-in, but OEM pre-installed versions of Windows might also have some other third-party Firewall, or is that really rare?

indirectlylit: okay, so just call it 'Windows Firewall'?

benjaoming: But then it would seem entirely irrelevant to non-Windows users.. "Firewall (Windows)" was intentional :)

indirectlylit: 'Firewall (Windows)' and 'Windows Firewall' both seem equally specific to Windows to me...

`¯\_(ツ)_/¯`

benjaoming: I removed the `(Windows)` part now, that way we're sure that all users will acknowledge the section, but you don't have to read more than a few sentences to understand that it's mainly a Windows thing.

indirectlylit:
coolio :) let's get this thing in

indirectlylit:
I do still think we should consider re-organizing a bit - e.g. https://github.com/learningequality/kolibri-docs/pull/50#pullrequestreview-156129432

radinamatic:
Gotcha @indirectlylit, will add on the 0.11 docs updates list! :+1:

benjaoming:
@indirectlylit I agree that we need to re-organize, but until then, I'm fine about just having a list of things we can quickly refer to.

The docs are portable and offline-friendly, so they should definitely be structured, easy to navigate (and read in a PDF) and contain information and not expect users to be able to ask questions online.

radinamatic:
> 1. Move the *Report a problem with Kolibri* information to the existing (and more visible) *Get in touch* page

Those two sections had almost identical content so I basically conserved just the one on the  *Get in touch* page.

> 2. Move the lengthy About IP addresses section to its own sub-page, or cut it way down in length

Going under the assumption that we would still want it to be under *Manage*, not sure what would be gained by separating it completely on its own page. Could moving it at the bottom of the *Troubleshooting* section, and adding a [simple link](https://github.com/learningequality/kolibri-docs/pull/54/files#diff-5f783e99ad465665334f124ff64d1627R12) inside the *Network issues* work well?

> 3. Rename the section to just 'Troubleshooting'

Done! :slightly_smiling_face:

benjaoming:
> Going under the assumption that we would still want it to be under Manage, not sure what would be gained by separating it completely on its own page. Could moving it at the bottom of the Troubleshooting section, and adding a simple link inside the Network issues work well?

Having longer tips and warnings at the bottom of pages works well IMO :+1: 

(Shorter tips and warnings or very important tips and warnings should go in-context)

Issue #33, opened 2018-07-09
raspberry pi instructions outdated


indirectlylit:
* [x] still references the `kolibri` user: https://github.com/learningequality/kolibri-docs/pull/32/files#diff-d3596120b107503957511384889d9a08R40

cc @benjaoming @radinamatic 

* [x] Also, we now have the 'change content directory' command. We should probably update the instructions to use that?

indirectlylit:
Also, this line is no longer true:

>  Loading channels can take a **long time** on a Raspberry Pi. When generating channel contens for Khan Academy, the step indicated as "Generating channel listing. This could take a few minutes…" could mean ~30 minutes

benjaoming:
True that! Thanks for finding this!

radinamatic:
Yes, the latest RPi improvements were added *after* we reviewed the RPi docs section last week... 🤦‍♀️

indirectlylit:
the inexorable march of progress!

indirectlylit:
While doing some basic rpi load testing today, we noticed that loading of exercises was slow.

As @jamalex said:

> One thing we didn't do in this setup (and not sure whether we recommend in our docs for the Pi) is put nginx in front of Kolibri to serve up static assets and help handle heavier concurrency

Also from @rtibbles:

> the zip endpoint loading seems to be a bottleneck, judging from the exercise data and image loading slow downs. So, adding nginx on the RPi would be important for helping with that under load

Sounds like it would be smart to include nginx instructions for rpi

benjaoming:
@indirectlylit this will be handled in a separate Debian package, AFAIK. Putting effort into making people do this through their own configurations would be a very temporary solution and would require to somehow tell people to revert those changes once that package arrives.

indirectlylit:
That's fair. Then maybe we should just add a note for advanced users? E.g. something like:

> Note that low-power devices like the Pi will not currently handle more than about 15 concurrent users. This performance will increase in future releases as we make optimizations and provide options to run Kolibri behind nginx. Advanced users may wish to experiment with setting up nginx themselves; however please be aware that we cannot provide support for this configuration and it may cause conflicts with a future update.

benjaoming:
That's a good note - will include that :)

indirectlylit:
Another recommendation reported here: https://github.com/learningequality/kolibri/issues/4040

When pip-installing on a Pi, use `--no-cache-dir` to prevent `MemoryError`s

[...]

PR #51, opened 2018-10-09
Updated Raspberry Pi instructions

benjaoming:
This addresses #33

I will retrieve my RPi for testing purposes tomorrow and run through all this manually again on a clean installation.

So the PR is not ready for testing instructions, but any kind of general ideas about the structure are welcome!

### DIFF
--- a/docs/install/_rpi.rst
+++ b/docs/install/_rpi.rst
@@ -7,13 +7,15 @@ There are several varieties of operating systems for Raspberry Pi. This guide is
 
 Kolibri is intended for **Raspberry Pi Model 3** and upwards.
 
-.. note:: The standard Raspbian choice has a graphical desktop. But you can also install **Raspbian Lite** that uses less resources but only has a command line interface. This documentation works seemlessly on both, but you might find that some packages are already installed in the desktop version of Raspbian.
+.. note:: The standard Raspbian OS has a graphical desktop. You can also install Raspbian Lite which uses fewer resources, but only has a command line interface. The instructions in this documentation work seamlessly on both.
 
 
 Install
 -------
 
-#. We need to upgrade the ``python3-cffi`` library, which is outdated on Raspbian. Upgrade it like this::
+#. We need to upgrade the ``python3-cffi`` library, which is outdated on Raspbian. Upgrade it like this:
+
+   .. code-block:: bash
 
       sudo apt install libffi-dev
       sudo pip3 install pip --upgrade
@@ -22,12 +24,14 @@ Install
 
   .. warning:: If you are installing Kolibri on the older **Raspbian Jessie** which is from before August 2017, you need to complete a few extra steps:
   
-     ::
+   .. code-block:: bash
     
         sudo apt install python3-pip python3-pkg-resources
         sudo pip3 install setuptools --upgrade
 
-#. Add our Ubuntu PPA with these special instructions::
+#. Add our Ubuntu PPA with these special instructions:
+
+   .. code-block:: bash
 
       sudo apt install dirmngr
       sudo su -c 'echo "deb http://ppa.launchpad.net/learningequality/kolibri/ubuntu xenial main" > /etc/apt/sources.list.d/learningequality-ubuntu-kolibri-xenial.list'
@@ -42,10 +46,12 @@ Install
 
   * **System time** isn't set properly or resets during power-off. This causes errors while downloading software. For instance, SSL certificates for online sources will fail to validate. Ensure that you have the right timezone in ``/etc/timezone`` and that the clock is set properly by running ``sudo ntpd -gq``.
 
-  * **Storage space** is often a scarcety. If you have a USB source for additional storage, you can use a ``kolibri`` management command or create your own symbolic links to have the data folder located elsewhere.
+  * **Storage space** is often scarce. If you have a USB source for additional storage, you can use a ``kolibri`` management command or create your own symbolic links to have the data folder located elsewhere.
   
-    Using the built-in management command::
-      
+    Using the built-in management command:
+
+    .. code-block:: bash
+
         # Stop kolibri
         sudo systemctl kolibri stop
         # Move the data
@@ -54,7 +60,9 @@ Install
         sudo systemctl kolibri start
         
   
-    Using symbolic links, you need to start and stop Kolibri and to set the permissions correctly::
+    **Or** using symbolic links, you need to start and stop Kolibri and to set the permissions correctly:
+
+    .. code-block:: bash
 
         # Stop kolibri
         sudo systemctl kolibri stop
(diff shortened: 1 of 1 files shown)


docs/install/_rpi.rst
```diff
@@ -1,52 +1,73 @@
...
+Raspberry Pi
+============
 
-Compatibility
--------------
+There are several varieties of operating systems for Raspberry Pi. This guide is intended for and tested on `Raspian <https://www.raspberrypi.org/>`__, the most popular choice of OS, based on Debian. To obtain and install Raspbian, refer to the official documentation.
+
+Kolibri is intended for **Raspberry Pi Model 3** and upwards.
+
+.. note:: The standard Raspbian choice has a graphical desktop. But you can also install **Raspbian Lite** that uses less resources but only has a command line interface. This documentation works seemlessly on both, but you might find that some packages are already installed in the desktop version of Raspbian.
```
indirectlylit: > The standard Raspbian OS has a graphical desktop. You can also install Raspbian Lite which uses fewer resources, but only has a command line interface. This documentation works seamlessly on both; however you might find that more packages come pre-installed in the desktop version of Raspbian.

docs/install/_rpi.rst
```diff
@@ -1,52 +1,73 @@
...
 
-Kolibri works well on Debian-based distributions for `Raspberry Pi <https://www.raspberrypi.org/>`_, such as Raspbian, and has been tested on RPi 3 models.
 
 Install
 -------
 
-1. Running ``add-apt-repository`` as shown in the PPA instructions does not work. Instead, run::
+#. We need to upgrade the ``python3-cffi`` library, which is outdated on Raspbian. Upgrade it like this::
```
indirectlylit: FYI I've been trying to migrate all uses of the `::` shorthand to more explicit:

```
.. code-block:: bash

      sudo apt install libffi-dev
      sudo pip3 install pip --upgrade
      sudo pip3 install cffi --upgrade
      sudo systemctl start kolibri
```

This enables syntax highlighting and also avoids syntax that originally confused me when learning RST

docs/install/_rpi.rst
```diff
@@ -1,52 +1,73 @@
...
+#. When the command finishes, open the default browser at http://127.0.0.1:8080 and proceed with the :ref:`setup_initial` of your facility. 
 
 
 .. note:: The following issues are quite common on a Raspberry Pi:
 
-  * The time isn't set properly and you will have errors downloading software. For instance, SSL certificates for online sources will fail to validate. Ensure that you have the right timezone in ``/etc/timezone`` and that the clock is set properly by running ``sudo ntpd -gq``.
+  * **System time** isn't set properly or resets during power-off. This causes errors while downloading software. For instance, SSL certificates for online sources will fail to validate. Ensure that you have the right timezone in ``/etc/timezone`` and that the clock is set properly by running ``sudo ntpd -gq``.
 
-  * You run out of storage space. If you have a USB source for additional storage, do something like this::
+  * **Storage space** is often a scarcety. If you have a USB source for additional storage, you can use a ``kolibri`` management command or create your own symbolic links to have the data folder located elsewhere.
```
indirectlylit: > **Storage space** is often scarce.

Can we remove the reference to creating symbolic links? I understand that people might do this and that's OK, but I think of any reason for us to officially support it.

docs/install/_rpi.rst
```diff
@@ -1,52 +1,73 @@
...
+        sudo ln -s /your/external/media/kolibri_data /var/kolibri/.kolibri
         # Start kolibri
-        sudo systemctl kolibri start  
+        sudo systemctl kolibri start
 
-  * Loading channels can take a **long time** on a Raspberry Pi. When generating channel contents for Khan Academy, the step indicated as "Generating channel listing. This could take a few minutes…" could mean ~30 minutes. The device's computation power is the bottleneck. You might get logged out while waiting, but this is harmless and the process will continue. Sit tight!
+  * **I/O operations are slow**, this means that a typical bottleneck on a Raspberry Pi is file transfer. Once Kolibri is up and running, this will not be a bottleneck, but while copying initial contents of several gigabytes, you will experience this. Both the SD card reader and the USB ports will limit you at 50-80MB/sec. From our experience, it doesn't matter much whether you are using the main SD card reader for storage or some media connected to your USB. But you might save time if you can connect the SD card of USB storage to another device with faster transfer speeds, especially if you are replicating from other installations.
```
indirectlylit: there are also huge differences in speeds between differences in SD cards

docs/install/_rpi.rst
benjaoming: I know how this works, I've intentionally skipped syntax highlighting for bash, because it was unhelpful. But let's be consistent, I'll use this.

docs/install/_rpi.rst
benjaoming: Didn't we have a recent episode where someone didn't like the python command? Assuming because it took away control? I think that the perfect compromise is to show both methods....

docs/install/_rpi.rst
benjaoming: Yup, will add that as a tip!

docs/install/_rpi.rst
benjaoming: Thanks that's more precise!

docs/install/_rpi.rst
indirectlylit: I think the recent episode was that the user didn't understand the python command.

I don't believe it's appropriate for us to show alternate methods of doing things that we have an official solution for, unless there are tradeoffs to each method that users need to be aware of.

What's the benefit?

docs/install/_rpi.rst
```diff
@@ -1,52 +1,83 @@
...
-      sudo pip3 install pip --upgrade
-      sudo pip3 install cffi --upgrade
-      sudo systemctl start kolibri
 
-3. When the command finishes, open the default browser at http://127.0.0.1:8080 and proceed with the :ref:`setup_initial` of your facility. 
+.. note:: The following issues are quite common on a Raspberry Pi:
+
+  * **System time** isn't set properly or resets during power-off. This causes errors while downloading software. For instance, SSL certificates for online sources will fail to validate. Ensure that you have the right timezone in ``/etc/timezone`` and that the clock is set properly by running ``sudo ntpd -gq``.
 
+  * **Storage space** is often a scarce. If you have a USB source for additional storage, you can use a ``kolibri`` management command or create your own symbolic links to have the data folder located elsewhere.
```
indirectlylit: is often scarce

:)

benjaoming: Argh, this is sloppy, sorry - was formatting SD cards and jumping back and forth on the KVM switch :) Fixed!

indirectlylit: better than "is often a farce"

benjaoming: :smile:

docs/install/_rpi.rst
benjaoming: Let's say I have this layout:

* SD Card: OS w/ kolibri installation and `~/.kolibri` (containing per-device user database, device registration etc)
* USB attached storage: I keep copying this media, it's bundled with content. I want to reuse them.

I've opened up #52, I will move these more detailed instructions and tips over there later on. I don't see much benefit in discussing the relevance of these detailed tips/instructions for the sake of maintaining simplicity... if the argument is that a detail is too nitty for the general level, then we'll never get to a point where the documentation (field manual) contains details.

I agree that the installation instructions should be simple -- I think we'll get there once the structures (such as a howto) are setup to move these couple of things over.

Would that work?

benjaoming:
I have tested these instructions on Raspbian Lite (Stretch) and they work on a clean installation. I made additional notes about setting up the system in a HOWTO proposal.

docs/install/_rpi.rst
benjaoming: Anyways, I moved it to the HOWTO now to ensure that this PR is ready :+1:

indirectlylit: I guess my point is that if we're presenting multiple options, we ought to describe what the pros and cons are - otherwise the user is left without guidance on the recommended way to do things. 

Kind of the zen of python maxim "There should be one-- and preferably only one --obvious way to do it."

If our built-in command works for most use-cases, we should recommend it. If it has limitations we should document the workarounds.

benjaoming:
Okay, I'm comfortable about merging this and moving on to specific feedback, we can track that in a new issue since #33 is already cluttered.

PR #52, opened 2018-10-11
Raspberry Pi Tutorial

benjaoming:
Opening this PR to flag contents of a tutorial.

It can be argued to be moved elsewhere, such as community forums. I would argue that we need to have a couple of solid field-ready guides for installing and using Kolibri on specific platforms. The Raspberry Pi is probably the most odd one out, because it will always be used by DIY people. So the intention of this guide is to give them a modular step-by-step guide to get started and understand how Kolibri integrates with their system.

Onwards, they can always choose other solutions or continue building their own. I would think that the community forums would benefit from having this as a resource to refer to.

I don't think we should finish and publish this Howto before we have a .deb that has Nginx and UWSGI integration set up. This will be a boost for smaller hardware platforms.

TODO:

 * [x] Figure out the ideal location of Tutorials and FAQ
 * [ ] Add additional resource links for already integrated systems -- IIAB, Kiwix, KoomBook, RACHEL etc.
 * [x] Add link to a thread about MicroSD quality/recommendations and benchmarks like http://www.pidramble.com/wiki/benchmarks/microsd-cards
 * [x] Finalize outline
 * [x] Finalize content
 * [ ] Test
 * [ ] Possibly, provide a non-supported "cheat" .img with Raspbian Lite pre-setup with Kolibri latest version from PPA.

image

image


benjaoming:
@radinamatic @indirectlylit how do you prefer to consistently write command line instructions, with or without the terminal indicator?

```
sudo apt update
```

or

```
$ sudo apt update
```

radinamatic:
Good question, I was reading a similar discussion recently in the `Write The Docs` newsletter - no clear best practice there... I find it repetitive and not extremely useful, especially for the tech-savvy user, and when there are several lines of commands. Command line text is already formatted sufficiently well that I don't find it necessary.

benjaoming:
@radinamatic agreed, it makes it more copy-paste friendly to omit the `$`. Let's do that. We can open a separate PR for a consequential change, I'll keep this PR just about the HOWTO.

indirectlylit:
agreed

benjaoming:
There is a special lexer for console snippets:

```rst
.. code-block:: console
   echo "mjello world"
```

It does not produce any special syntax, it's the same as using `bash`, but there is a CSS class `.highlight-console` which we could potentially use to add our own styling. I don't think it's that critical :)

radinamatic:
Good to know, thanks for the tip! :wink:

benjaoming:
@jredrejo - in case you have time, I would love to read your inputs to these steps. Do you find anything problematic here? Btw. do you have a tip of which local domain name to use? I have put `kolibri.library`, in order to cut short and not talk about host names, custom DNS resolutions etc.. everything is just one big catch-all, which makes it very easy to create the captive portal.

For the review process, I suggest to render the documentation firstly, as it becomes a lot more readable :)

benjaoming:
@radinamatic I also would like to add this to 0.11 docs, because we might (optimistically speaking) be able to add some new work in the .deb packages for UWSGI and Nginx support that affectus the tutorial in 0.12... in this regard, I would like to get this guide published to inform that work and the following documentation updates. Simply because we can end up at an interesting outcome, where we much more actively encourage people to deploy Kolibri with a specific set of instructions.

radinamatic:
As long as everything works as described, I see no problem in adding this to 0.11 docs :wink:

benjaoming:
Here's a rendering of the current version: https://overtag.dk/rpi/install/tutorials/rpi.html

docs/tutorials/rpi.rst
```diff
@@ -0,0 +1,359 @@
+.. _tutorial_rpi:
+
+An offline Raspberry Pi server
+==============================
+
+This guide shows you how to configure a Raspberry Pi as a local WIFI hotspot serving Kolibri.
```
indirectlylit: Looks like `Wi-Fi`, `Wifi`, `wifi`, or `WiFi` are most common: https://english.stackexchange.com/questions/41180/how-do-you-spell-wifi-wi-fi-wifi

One idea from that thread I like is to call it a "wireless LAN" or WLAN instead.

indirectlylit (commented):
This looks very well-written.

Recommend we have someone from imps go through and actually try it before publishing as a 'QA' step.

docs/tutorials/rpi.rst
benjaoming: Thanks, fixed :)

benjaoming:
@indirectlylit Radina will do that -- she has a Pi that's eagerly awaiting installation :)

I'll try and do it myself before hand, as I hope to catch any obvious mistakes myself before handing it over. But I did try things out a couple of times.. as a final rehearsal.

benjaoming:
I have moved Tutorials to live in the "Install" section. The problematic properties of the drafted tutorials are that:

1. They are related to "admin users"
2. They are related to installation

Therefore, they sit badly in the top level, as these tutorials are not intended for Coaches or Learners. I also think that the rest of the documentation is written in a way that's quite close to being a tutorial, so the terminology would get lost.

So in order to give these tutorials proper scoping, let's assume that they are all install-related tutorials: Setting up Kolibri on various platforms and with various integration.

[edit: screenshot moved to description ^]

indirectlylit:
mind updating your demo? https://overtag.dk/rpi/install/tutorials/rpi.html

I'm a little confused by the screenshots in the description

benjaoming:
Oh yeah, done, thanks for the reminder :)

benjaoming:
Quick note that it relocated: https://overtag.dk/rpi/install/tutorials/rpi.html

docs/install/tutorials/rpi.rst
```diff
@@ -0,0 +1,379 @@
...
+The following commands work on Linux/macOS for setting up the .img files provided. You will also need to know the *device node* for the SD Card reader. On Linux, this is typically ``/dev/mmcblk0`` for the 0th card in your card reader.
+
+.. code-block:: console
+
+  # Unpack the .zip into memory and write it to <device node>
+  unzip -p /path/to/raspbian-stretch-lite.zip | sudo dd of=/dev/mmcblk123 bs=4M conv=fsync
+
+.. tip:: Read the official guides for setting up your card: `Copying .img files <https://www.raspberrypi.org/documentation/installation/installing-images/README.md>`__
+
+.. tip:: Select your MicroSD card based on other people's experiences and technical benchmarks. For more information, read `this Community Forums thread <http://community.learningequality.org/t/microsd-cards-picking-the-right-one-experiences-and-benchmarks/935>`__.
```
benjaoming, line 55: This is the tip that was added based on @jredrejo 's feedback

indirectlylit:
Thanks for updating!

Regarding the organization, a few thoughts:


| 'install' top-level | raspberry pi tutorial | multi-core "tutorial" |
|--|--|--|
| image | image | image |

* We only have one tutorial right now: raspberry pi. Preemptively adding the uWSGI placeholder before it's written is misleading: readers will be disappointed by what they find.
* Navigationally, it's a confusing that all the links under 'Install Kolibri' point to anchors within `install.html` except 'Tutorials', which points to a nested tree structure. I would expect all links at the same level in the TOC to have similar behavior
* The raspberry pi tutorial currently appears above more general installation-related information in the TOC
* As a reader I'm left confused about the difference between the 'Raspberry Pi' section of 'install' and the new tutorial - looks like they overlap a bit?

A possible solution to all of these would be to remove the raspberry pi tutorial from the TOC, and instead put a very prominent link to it inside the Raspberry Pi section of the Install page. This would help people who need it to find it more easily because it's associated with "raspberry pi installation" instead of hidden in a more general-sounding "tutorials" section:

| Current | Possible alternative |
|--|--|
| image | image |

benjaoming:
The trade-off is about permalinks. If moving the tutorial, as I expect we will/should, then we should not link to it once it's published... or update those links, which I think isn't realistic. I can remove the uWSGI placeholder, but I took advantage and used it to put important information about its outcomes.

This PR is for `develop`, so intended for 0.12.. and I think we might have the Debian package w/ uWSGI and Nginx configuration ready by then...

indirectlylit:
>  I think we might have the Debian package w/ uWSGI and Nginx configuration ready by then...

good point. So in that case presumably we won't need a tutorial?

indirectlylit:
Regarding permalinks, I think the solution is to always link to the release-versioned URLs.

That way we get the best of both worlds:

* flexibility to re-arrange docs between releases
* links that never break (and will be properly labeled because of your work in https://github.com/rtfd/readthedocs.org/pull/4842)

benjaoming:
> good point. So in that case presumably we won't need a tutorial?

I think we still do. Simply because the package will require separate installation instructions one way or the other. I can't tell for sure, but this part (because of the strings to non-Debian platforms) could be a tutorial.. it could also be lots of new sections for the installation instructions. I think we will want to add a bit of discussion such that users understand that these settings can be tweaked with regards to 1) cpu cores and 2) memory. Not sure if there's much else to it. But as of now, I see it as a tutorial, not a 1-2-3 install guide.

indirectlylit:
well in any case, it's beside the point because @radinamatic hasn't written it yet. Let's avoid adding new structure until she decides is necessary

benjaoming:
@indirectlylit how can I avoid adding new structure?

This is a tutorial, not an install guide.. if we put it in the "Install" section as the official/default Raspberry Pi installation instructions, it's going to be quite heavy for a new user to behold :)

indirectlylit:
please take another look at the 'Possible alternative' I proposed above. I was trying to find a balance that makes it easy to find for pi users, keeps the TOC consistent, and avoids adding new unnecessary structure.

benjaoming:
I've moved the uWSGI section into a draft file and removed it from the TOC. I can expand it step by step as we approach what the classroom servers really mean, what the install process is, and which considerations we would find beneficial to communicate through a tutorial.

jredrejo:
> I've moved the uWSGI section into a draft file and removed it from the TOC

great, I think that should be superseded by the kolibri-server info

benjaoming:
@jredrejo - I agree, but also renamed the title of the draft page to "Classroom server", maybe this theme/scope has a better potential for supplementing the new package :)

benjaoming:
@indirectlylit I have spent quite some time trying to include the tutorial without displaying it in the TOC, and I can't get it working. I don't want to spend more time on this, since creating a page outside of the TOC already seems like a bad idea.

Curious: How did you get the "Important note" box?

indirectlylit:
> I have spent quite some time trying to include the tutorial without displaying it in the TOC, and I can't get it working.

You do it using the `:orphan:` directive at the top of the tutorial file. Cool feature - sphinx actually warns that it's not referenced from the TOC.

> creating a page outside of the TOC already seems like a bad idea

In general I definitely agree, but in this case I believe it's the right move for the reasons I explained above.

> Curious: How did you get the "Important note" box?

I may have cheated and hacked that with chrome dev tools :)

radinamatic:
Awesome tute @benjaoming!!! :100: 

I would agree with @indirectlylit's [above concerns](https://github.com/learningequality/kolibri-docs/pull/52#issuecomment-448029902), and recommend we keep it out of the TOC for now (the `:orphan:` directive works like magic :heart:).

One thing I found is that the long figure caption overlaps the tip below, and behaves strangely when changing the browser width. 

figcaption

I recommend shortening the caption and putting the content in the regular paragraph below the pic.

ledev1901 start running - oracle vm virtualbox_413

benjaoming:
@radinamatic you will enjoy this PR :smile: 

https://github.com/rtfd/sphinx_rtd_theme/pull/723

docs/install/tutorials/uwsgi.rst.draft
```diff
@@ -0,0 +1,15 @@
+.. _how_to_uwsgi:
```
jredrejo, line 1: I think we could replace this section by advicing the installation of kolibri-server. If the instructions to install kolibri are followed, installing kolibri-server is as simple as 
`sudo apt-get install kolibri-server`
The package is now in the kolibri-proposal ppa but will be in the kolibri ppa as soon as 0.12 is released.

Obviously these instructions are valid, but quite more complex.

jredrejo (changes requested):

docs/install/tutorials/uwsgi.rst.draft
benjaoming: This was in a `.draft` file, so not actually in the docs.

indirectlylit: If it's a draft that will be superseded anyway, might make more sense to just remove it?

benjaoming: It contained some text that I feel could be reused.

PR #62, opened 2019-01-21
merge v0.10.x -> v0.11.x

benjaoming:
I made a change to 0.10.x that I wanted to quickly merge in, but what happened here?

I know that these are my changes, but I need to grasp properly that the changes are okay with you @radinamatic  and @indirectlylit - also, I will get back with a small report on how this builds.


indirectlylit:
Most of these changes look related to restructuring the location of images using the `figure_language_filename` variable.

Most of the image diffs look like renames without modification. However I would recommend confirming that, because we want to make sure not to overwrite an 0.11 screenshot with an 0.10 screenshot

benjaoming:
ping @radinamatic

radinamatic:
Built & tested locally as recommended, looking good! :+1: :
```
git fetch upstream
git checkout upstream/release-v0.11.x
git merge upstream/release-v0.10.x
```

radinamatic (approved):
LGTM! :100:

PR #69, opened 2019-03-18
New gettext messages gen'ed

benjaoming:


benjaoming:
This is auto-generated, can't see the reason to review.

PR #70, opened 2019-03-23
Minor Typo: Raspbian

arky:
The correct spelling is "Raspbian" and the project address is 
http://raspbian.org/. 

Alternatively you also use the following URI as well https://www.raspberrypi.org/downloads/


radinamatic (approved):
Thank you @arky!

LGTM!

PR #74, opened 2019-04-02
First stab at changing PDF output

### LINKED ISSUE #23: Improve the PDF output from ReadTheDocs platform
Current [default PDF output from the RTD](https://media.readthedocs.org/pdf/kolibri/latest/kolibri.pdf) version of the docs is not optimal and it can use several improvements.

- better text & page formatting
- consistency in image sizes
- accessibility

radinamatic:
(I hope it eventually) Fixes #23 
So much still to learn about LaTeX 😩 

[kolibri.log](https://github.com/learningequality/kolibri-docs/files/3033336/kolibri.log)
[kolibri.pdf](https://github.com/learningequality/kolibri-docs/files/3033337/kolibri.pdf)


indirectlylit:
I'm curious whether using [rst2pdf rather than using latex](https://docs.typo3.org/typo3cms/extensions/sphinx/AdvancedUsersManual/RenderingPdf/Index.html) might be a faster path to success for our current needs? The [docs assert that](https://docs.typo3.org/typo3cms/extensions/sphinx/AdvancedUsersManual/RenderingPdf/LaTeXVsRst2pdf.html) latex gives a "much prettier and professional" output, but that will only be true after hours of fine-tuning it (or maybe a pro like @ivanistheone helping out).

At the very least, I think having it look similar to our online docs (e.g sans-serif font) would actually be a good thing for continuity and recognizability.

PR #78, opened 2019-04-06
Newest latexpdf

ivanistheone:
Same as https://github.com/learningequality/kolibri-docs/pull/74 with more fixes

Summary of changes:
 - Set header to be:
    `{{section_counter}}  {{section_title}}                                     {{page_number}}`
      - in italics to distinguish from main flow
      - small so fits even longest section title
         Screen Shot 2019-04-11 at 11 08 01 AM
      - no footer
  - Use Lato font
  - Use a4paper (most likely outside of US)
  - made external hyperlink targets show up as footnotes


ivanistheone:
Lates PDF preview: [kolibri_774357c.pdf](https://github.com/learningequality/kolibri-docs/files/3069264/kolibri_774357c.pdf)

Thinks left TODO:
  - further tweaks to code envs (apparently not a lot of customizability so may have to live with the current "old school" font)
  - set PDF keywords and subject
Screen Shot 2019-04-11 at 11 00 58 AM

@radinamatic I'm handing over this branch for you to play with it, and check if builds OK on  rtd

docs/conf.py
```diff
@@ -26,6 +26,7 @@
 
 extensions = [
     'sphinx.ext.todo',
+    'sphinx.ext.autodoc',
```
ivanistheone, line 29: We probably don't need this since not using any text from docstrings...

docs/conf.py
```diff
@@ -234,3 +245,217 @@ def setup(app):
...
+        \renewcommand{\sectionmark}[1]{\markboth{}{\emph{\thesection~#1}}}
+        %%%% optional: uncomment next line to change header on every \section and not on \subsection
+        %%%% \renewcommand{\subsectionmark}[1]{}
+
+
+        %%reduce spacing for itemize ???
+        \usepackage{enumitem}
+        \setlist{nosep}
+
+        %%%%%%%%%%% Quote Styles at the top of chapter ???
```
ivanistheone, line 331: Can probably remove this and next five lines since we don't use

docs/conf.py
```diff
@@ -234,3 +245,217 @@ def setup(app):
...
+        %%%  COVER PAGE  %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+        \pagenumbering{Roman} %%% to avoid page 1 conflict with actual page 1
+
+        \begin{titlepage}
+            \centering
+
+            \vspace*{80mm} %%% * is used to give space from top
+            \begin{figure}[!h]
+            \centering
+            \includegraphics[scale=0.75]{logo.png}
```
ivanistheone, line 369: This would make a better cover-page logo since it shows contexts of use:
https://github.com/learningequality/kolibri-docs/blob/develop/docs/img/kolibri-use-cases.png

docs/conf.py
```diff
@@ -234,3 +245,217 @@ def setup(app):
...
+     u'Kolibri User Guide',                 # title,
+     u'Published by learningequality.org',  # author,
+      'manual'                              # documentclass [howto, manual, or own class])
+     ),
+]
+
+# The name of an image file (relative to this directory) to place at the top of
+# the title page.
+#latex_logo = None
+latex_logo = 'logo.png'
```
ivanistheone, line 442: this is probably not used since we do custom cover-page in `maketitle`

docs/conf.py
indirectlylit: @khangmach do we have an SVG copy of this?

khangmach: No, there's only a PNG format at the moment. Do we need an svg?

indirectlylit: if available, it would come out nicer when printed

indirectlylit: Doesn't need to be this either. I think we're just choosing an illustration to put on the front

docs/conf.py
radinamatic: True, may remove eventually!

docs/conf.py
radinamatic: That's precisely what I wanted to use on the cover page @khangmach (see the sample doc above), unless you have a better suggestion...?

In any case, SVG version of whatever we decide to use would be great to have!

docs/conf.py
radinamatic: I think that's what specifies which file gets displayed in the cover page, right?

radinamatic (approved):
Awesome progress, approving so we can test on RTD and continue tweaking!

docs/conf.py
khangmach: yeah i'll spin up an SVG version of that illustration

khangmach: should I forward it to you on slack or

indirectlylit: cool, maybe post in the `#documentation` room

docs/conf.py
ivanistheone: If we were to use the default cover page (provided by basic sphinx class) then it would use the logo specified here, but since we define a custom cover page,
https://github.com/learningequality/kolibri-docs/blob/develop/docs/conf.py#L339-L390
the `latex_logo` option is not used.

@radinamatic  If you want to change the cover image change this line: https://github.com/learningequality/kolibri-docs/blob/develop/docs/conf.py#L369
replacing with some high resolution pdf of the `.svg` that Khang sent

PR #81, opened 2019-04-12
added high-res PDF for coverpage

ivanistheone:
opened `.svg` using AffinityStudio and exported as PDF

@radinamatic  I was wrong about what I said earlier that `latex_logo` not being used --- it's necessary to have the same filename there too so that it gets copied to the `_build/latex/` directory...


benjaoming:
Thanks @ivanistheone - we should get this in, can we retarget for 0.12?

ivanistheone:
@benjaoming This PR is a patch on top of the new pdf styling commits from 2019-04-05 and 2019-04-11 which are currently on develop, so doesn't make sense to re-target to 0.12.x right now.

benjaoming:
Don't see anything holding back this improvement.

I do not have the rather hefty LaTeX libraries locally, so am merging this to see if it builds / displays fine in RTD and will feed back here.

benjaoming:
Looks great! https://kolibri.readthedocs.io/_/downloads/en/develop/pdf/

PR #87, opened 2019-05-04
Updates for 0.12.3 release

### LINKED ISSUE #55: audit usage of 'setup' vs 'set up'
e.g. "Check your setup for issues" (noun) vs "Set up the account" (verb)

### LINKED ISSUE #60: Separate instructions for Debian
Installing from .deb from the PPA on Debian requires roughly the following instructions on Debian 9:

```
sudo apt-get install software-properties-common dirmngr
sudo su -c 'echo "deb http://ppa.launchpad.net/learningequality/kolibri/ubuntu cosmic main" > /etc/apt/sources.list.d/learningequality-ubuntu-kolibri-cosmic.list'
sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys DC5BAA93F9E4AE4F0411F97C74F88ADB3194DD81
sudo apt-get update
sudo apt-get install kolibri
```

Since instructions are sufficiently different from the default Ubuntu install steps, it might be worth having the separate section for Debain install instructions to guarantee optimal copy-and-pastability for successful install.

### LINKED ISSUE #79: [Docs] Kolibri Server Configuration
Please kindly add some documentation on settings of your kolibri server package using the various configuration files in /etc/default /etc/kolibri etc.

This information is useful for installations on Gnu/Linux distributions and also on ARM based devices like Raspberry Pi.

### LINKED ISSUE #80: Recommend Simpler Captive Portal default domain 'kolibri.library'
In my deployment I found that typing 'http://kolibri.library' was too long and difficult for non-native speakers on tiny mobile keyboards. I reverted to using simpler 'http://box/' 'http://box.lan' naming scheme for my Kolibri deployments. 

This is much easier for our users. Both 'Kolibri' and 'library' have prone to mis-spellings. Also OpenWRT by default recommends using '.lan' domain for its local domains, instead of '.local' to avoid conflict with ZeroConf. 

Of course, as a fallback I have standard captive portal dns redirects in place. Hope this feedback would make the kolibri docs better!

### LINKED ISSUE #85: user confused by command line instructions
In the instructions, we say 

> To export Kolibri content channels on a local drive in order to share it with another device, run these two commands in sequence. The first exports the channel database, and the second exports the resources (videos, documents, etc.).
> 
> ```
> kolibri manage exportchannel -- <Channel ID> /path/to/local/drive/KOLIBRI_DATA
> kolibri manage exportcontent -- <Channel ID> /mount/mydrive/KOLIBRI_DATA
> ```
> 
> The path should be to a folder named `KOLIBRI_DATA` at the root of the local drive, so it will get picked up later for importing via the Web UI.

One user thought they were supposed to keep the `<` `>` and tried

```
kolibri manage exportchannel -- <ec164fee25ee526296e68f71e169> /Volumes/mydrive/KOLIBRI_DATA

### PR DESCRIPTION
**Pull request by radinamatic**
Ready for review!

- [X]  fixes #55 (Setup vs. Set up)
- [X]  fixes #80 (Captive portal domain)
- [X]  fixes #85 (import channel command line improvements)
- [X]  fixes #79 (documents the pre-release of ``kolibri-server``)
- [X]  fixes #60 (Move all install instructions for RPi in the Tutorials section)
- [X]  documents changes in ``dbrestore`` command [#5344](https://github.com/learningequality/kolibri/pull/5344)

### CODE REVIEW

File: docs/install/_linux.rst [hunk 1]
```diff
@@ -23,6 +23,32 @@ Use the following commands in `Terminal <https://help.ubuntu.com/community/Using
...
+To install **kolibri-server** follow these steps.
+
+#. Run these commands in the Terminal:
+
+	.. code-block:: bash
+
+		sudo apt-get install software-properties-common python-software-properties
+		sudo add-apt-repository ppa:learningequality/kolibri-proposed
+		sudo apt-get update
+		sudo apt-get install kolibri-server
```
benjaoming, line 63: I think we should add a warning that this also switches the setup to use the latest built Kolibri pre-release?

docs/install/tutorials/rpi.rst
```diff
@@ -251,11 +251,11 @@ You can use :guilabel:`CTRL` + :guilabel:`SHIFT` + :guilabel:`V` to paste text i
 Installing Kolibri
 ------------------
 
-**Firstly**, follow the main instructions for installing :ref:`Kolibri on Raspberry Pi <rpi>`.
+**First**, follow the main instructions for installing :ref:`Kolibri on Raspberry Pi <rpi>`.
 
-After completing the installation, you can make kolibri available on port ``:80`` in addition to ``:8080``. This will make it possible to type ``kolibri.library`` in the browser location bar, and because of our captive portal, it will display.  
+After completing the installation, you can make kolibri available on port ``:80`` in addition to ``:8080``. This will make it possible to type, for example, a domain ``box.lan`` in the browser location bar, and because of our captive portal, it will display.
```
benjaoming: Isn't `box.lan` the default of IIAB? I understand that `library` can be difficult to spell, maybe `kolibri.lan` as a compromise?

benjaoming (approved):
These are great changes, left 2 comments :)

docs/install/tutorials/rpi.rst
indirectlylit: Couldn't it be simply `kolibri` (similar to `localhost`)?

I don't think a TLD is actually necessary, and I see some indication that [it should be avoided](https://serverfault.com/questions/17255/top-level-domain-domain-suffix-for-private-network).

benjaoming: It's to avoid a browser going for its default search engine. Typing just "kolibri" will create a Google search, rather than going to the DNS=>captive portal. I've tested it.

indirectlylit: true. you would have to type `http://kolibri` which is not good

docs/install/_linux.rst
radinamatic: Righto, will add! :+1:

docs/install/tutorials/rpi.rst
radinamatic: Ok, will leave `kolibri.lan` as our default, and add a tip that the domain name can be set up to a different value according to the demands of their specific implementation.

docs/install/_linux.rst
```diff
@@ -32,6 +32,42 @@ Install from PPA repository
...
 
+Uninstall
+*********
+From the command line: ``sudo apt-get remove kolibri``.
+
+Upgrade
+*******
+
+When you use the PPA installation method, upgrades to newer versions will be automatic, provided there is internet access available.
+
```
radinamatic, line 43: @jredrejo @benjaoming Do you think this location is the best for documenting ``kolibri-server``? it is a bit _hidden_ in the TOC, which might not be bad, given that it is still a pre-release...

Another thing to decide is if we want to add more detailed instructions as requested in #79.

docs/install/tutorials/rpi.rst
```diff
@@ -251,11 +253,57 @@ You can use :guilabel:`CTRL` + :guilabel:`SHIFT` + :guilabel:`V` to paste text i
 Installing Kolibri
 ------------------
```
radinamatic, line 255: @benjaoming Moved all the RPi installation instructions here, but I'm still unsure if we are keeping the ``cffi`` library or not...?

radinamatic:
@benjaoming @jredrejo @indirectlylit 

I think all the changes that should go into the 0.12.3 update are in, let me know if something is still missing or needs further amends.

docs/install/tutorials/rpi.rst
benjaoming: I talked to @jredrejo about it, and I think it's a flaky/flimpsy build error that I've been testing with recently, where possibly the armv6 `cffi` library bundled with our C extensions got left out. But I cannot test it or verify it any further.

benjaoming: So if you want a quick conclusion: Yes, remove it.

benjaoming: Also because _it's supposed to work_ without this instruction anyways!

docs/install/_linux.rst
benjaoming: Great point, yes, it will end up too hidden. As I might have mentioned before, I would like if we worked on a tutorial similar to the Raspberry Pi tutorial, perhaps unifying some concepts later on. I think the tutorial could be called "Class-room servers". _That's only for the long run_.

I think that same title could work well in the install sub-TOC directly below all the other items. It will have to mention loud and clear that it's only available on Debian/Ubuntu.

image

I'd suggest to create a new file `install/_server.rst` ?

```
Classroom servers
-----------------

.. tip: Using ``kolibri-server`` on a CPU with 4 cores means serving up to 4x as many users as you would on a regular Kolibri. This is because it adds a configuration to the system that runs many Kolibri workers in parallel.

Kolibri can be optimized further, and we recommend to do so. **This method is only available for Debian/Ubuntu** platforms, and you can find slightly more elaborated instructions for Raspberrry Pi [link].

As an addition to the main Kolibri packages, you can install ``kolibri-server``, a package for Debian/ubuntu systems that contain the configurations necessary to serve Kolibri in an optimized fashion for CPUs with several cores.

.. warming: This is a new distribution package introduced
```

benjaoming:
Hi @radinamatic - When doing PRs, it's much easier if you merge existing work that's already reviewed and open a new PR with new changes :heart:

docs/install/_linux.rst
benjaoming: Have added a suggestion for a re-location of this and a different heading further up. Basically, I don't think that `kolibri-server (pre-release)` is a very telling name :)

docs/install/_linux.rst
```diff
@@ -32,6 +32,42 @@ Install from PPA repository
...
+		sudo add-apt-repository ppa:learningequality/kolibri-proposed
+		sudo apt-get update
+		sudo apt-get install kolibri-server
+
+   During the installation you will be offered the option to choose the port (8000, 80 or leave the default 8080).
+
+#. Restart the system.
+
+.. warning:: * Be advised that this procedure also switches the setup to use the latest built Kolibri pre-release.
+	* Keep in mind that the **kolibri-server** system performance will depend not only on server device features, but on local WiFi access point characteristics.
```
benjaoming, line 70:
```suggestion
	* Keep in mind that Kolibri's performance in a user's browser will depend not only on server capacity and load, but also on the speed and capacity of the local network, for example your local WiFi access point.
```

I think we could consider linking to the system requirements section and expand it with some facts about kolibri-server as we go along :)

benjaoming (approved):
Had some few more suggestions, I can also merge this and open up a new PR suggesting the `_server.rst` refactor, if you like?

radinamatic:
> it's much easier if you merge existing work that's already reviewed

Yes, I can totally understand that, but please _don't_ review/approve a PR if it is still labeled as `WIP` :wink: 

I'm going to merge this now, and please _do_ open another PR with the proposal you suggested here! :+1:

benjaoming:
Thanks @radinamatic - I thought the WIP was removed before my review, but nevermind. I'll open a PR with the suggestion for moving the RPi instructions. Smiles :)

PR #88, opened 2019-05-04
0.11.x into 0.12.x

benjaoming:
Since the 0.12.x branch is now default, this is a bit urgent..

LMK about things that need manual changing in these merges.

There were no conflicts, except that 14.04 was changed to 16.04 in a single paragraph, and that was resolved to 16.04.


benjaoming:
Note that Sphinx 2.0 was released, we were previously on 1.8. I couldn't reproduce the warning message locally, so I'm forcing a Sphinx downgrade to see if it works.

benjaoming:
...update: it works to downgrade Sphinx.

docs/install/_linux.rst
```diff
@@ -6,23 +6,32 @@ Debian/Ubuntu
...
+    sudo apt-get install software-properties-common dirmngr
     sudo add-apt-repository ppa:learningequality/kolibri
     sudo apt-get update
     sudo apt-get install kolibri
 
+**In Debian-based distributions** you need to use these commands to point to our Launchpad PPA:
+
+.. code-block:: bash
+
+    sudo apt-get install software-properties-common dirmngr
```
jredrejo, line 29: I don't understand why this is needed if later the repository info is added manually. Either `software-properties-common` is needed to use `add-apt-repository` or it's not needed if the `/etc/apt/sources.list.d/*.list`file is created manually. 
Both things are not needed.

jredrejo (dismissed):

docs/install/_linux.rst
benjaoming: You are right. This was just intended as a merge PR, but we could fix later. I think `dirmngr` is still needed for `apt-key`.

benjaoming: Fixed in 5b4076c2ce8c72430bd69c3da8c1be0b9798a3c6

jredrejo: > You are right. This was just intended as a merge PR, but we could fix later. I think `dirmngr` is still needed for `apt-key`.

Yes, it does, if `apt-key` is used, but not if `add-apt-repository ppa:learningequality/kolibri` is used.
I prefer to use apt-key as it's compatible for Raspberry, we don't have a ppa for RPi, so add-apt-repository won't work for this device.

radinamatic (approved):
Having less experience with RPi than @jredrejo and @benjaoming, I cannot offer an opinion on the above discussion, but the rest of the changes seem OK to me, so will approve.

benjaoming:
Will generate new translation messages and push to CrowdIn and ping you back @radinamatic .

PR #89, opened 2019-05-04
Update translation messages in 0.11.x

benjaoming:
Also added a little fix for @jredrejo review in #88


benjaoming:
I didn't count the number of actual text changes, but there are very few. Most here is just automatic line numbering.

PR #112, opened 2019-07-17
kolibri-server in the TOC

### LINKED ISSUE #108: kolibri-server is not easily found in TOC
The `kolibri-server` section is not easily found in the table of contents because of the deep nesting:

image

https://kolibri.readthedocs.io/en/latest/install.html#kolibri-server-pre-release

Recommend making the heading one level higher, so it's a direct child of Debian/Ubuntu

radinamatic:
Fixes #108

### DIFF
--- a/docs/install/_linux.rst
+++ b/docs/install/_linux.rst
@@ -41,10 +41,10 @@ Upgrade
 
 When you use the PPA installation method, upgrades to newer versions will be automatic, provided there is internet access available.
 
-kolibri-server (pre-release)
-----------------------------
+kolibri-server
+--------------
 
-We are finalizing the release of an optimized version of Kolibri for Linux systems called **kolibri-server**. This version is recommended for:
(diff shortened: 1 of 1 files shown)


docs/install/_linux.rst
```diff
@@ -42,7 +42,7 @@ Upgrade
 When you use the PPA installation method, upgrades to newer versions will be automatic, provided there is internet access available.
 
 kolibri-server (pre-release)
-****************************
+----------------------------
```
benjaoming: @jredrejo what would you say if we removed "(pre-release)"? Thinking that the latest feedback has been encouraging. You still need the cache stuff to go in?

radinamatic: I'm good with removing this if you agree, will amend asap! :+1:

benjaoming: Yeah, I would also defend that decision, simply to get a gradually wider adaptation and keeping the huge benefits of this package in mind.

benjaoming (approved):
Smart simple impro!

PR #113, opened 2019-07-18
Slideshow docs

nucleogenesis:
This provides user docs for the new slideshow feature.

Please let me know if there are any other docs that I should add!


docs/learn.rst
```diff
@@ -260,6 +261,15 @@ To read digital books in Kolibri you have several available controls inside the
 
 .. _lessons:
 
+Slideshow Viewer options
```
benjaoming, line 261: You would have to move it up a bit because of the `lessons` label declaration now pointing to the slideshow header.

radinamatic: As @benjaoming noted, either move the `.. lessons:` reference so it immediately precedes the `Lessons` heading, or move the whole slideshow section above `.. lessons:`.

benjaoming: Haha, @radinamatic - I'm LOL'ing out loud right now :)

docs/learn.rst
```diff
@@ -260,6 +261,15 @@ To read digital books in Kolibri you have several available controls inside the
...
+""""""""""""""""""""""""
+
+To navigate the slides of a slideshow, you may use the on-screen navigation controls or your keyboard's arrow keys.
+
+* Use the |fullscreen| button in the upper right corner to open the ebook in fullscreen view.
+* With the :guilabel:`Left Arrow` or :guilabel:`Right Arrow` keys or the buttons on the side of the slideshow.
```
radinamatic: We use `:guilabel` to reference UI elements (buttons mostly) in Kolibri interface, and less often keyboard keys. Since these keys are not "spelled" out in words on the keyboard, and I don't think we can put the icon inside that directive, I'm hesitant if they should be used as-is. 

On another hand, we do have substitutions for `|next-page|` and `|previous-page|` used for ebook content, so we could reuse those or make a version styled as in the slideshow.

None of the above is a blocker, so feel free to pass, and I'll correct/add next week with the 0.12.6 update.

radinamatic (changes requested):
Just one small change required, otherwise LGTM! :+1:

docs/learn.rst
```diff
@@ -260,6 +261,15 @@ To read digital books in Kolibri you have several available controls inside the
 
 .. _lessons:
```
benjaoming:
```suggestion
```

docs/learn.rst
```diff
@@ -260,6 +261,15 @@ To read digital books in Kolibri you have several available controls inside the
...
 
+Slideshow Viewer options
+""""""""""""""""""""""""
+
+To navigate the slides of a slideshow, you may use the on-screen navigation controls or your keyboard's arrow keys.
+
+* Use the |fullscreen| button in the upper right corner to open the ebook in fullscreen view.
+* With the :guilabel:`Left Arrow` or :guilabel:`Right Arrow` keys or the buttons on the side of the slideshow.
+* You can choose a specific slide by clicking the round icons at the bottom of the slideshow view. 
+
```
benjaoming, line 269:
```suggestion
 
 .. _lessons:
 
```

docs/learn.rst
```diff
@@ -260,6 +261,15 @@ To read digital books in Kolibri you have several available controls inside the
 
 .. _lessons:
 
+Slideshow Viewer options
+""""""""""""""""""""""""
+
+To navigate the slides of a slideshow, you may use the on-screen navigation controls or your keyboard's arrow keys.
+
+* Use the |fullscreen| button in the upper right corner to open the ebook in fullscreen view.
```
benjaoming:
```suggestion
* Use the :guilabel:`|fullscreen|` button in the upper right corner to open the ebook in fullscreen view.
```

docs/learn.rst
benjaoming:
```suggestion
* With the Left Arrow (←) or Right Arrow (→) keys or the buttons on the side of the slideshow.
```

nucleogenesis:
@radinamatic @benjaoming Thank you for the reviews! I pushed changes according to your feedback - please let me know if I missed something.

radinamatic (approved):
LGTM, thanks @nucleogenesis! :+1:

PR #106, opened 2019-06-25
Document the procedure to update channels and import more content

radinamatic:
Ready for review!

Fixes #57

Issue #104, opened 2019-06-17
update raspberry pi tutorial to use kolibri-server package


indirectlylit:
The [Debian/Ubuntu section](https://kolibri.readthedocs.io/en/latest/install.html#kolibri-server-pre-release) says that

>  Typical use case would be a Raspberry Pi 3+ device with 4 CPU cores, which will perform much better and faster with kolibri-server compared to base Kolibri.

However the [RPi tutorial](https://kolibri.readthedocs.io/en/latest/install/tutorials/rpi.html) contains manual NGINX configuration instructions and does not mention `kolibri-server`.

indirectlylit:
refs: https://github.com/learningequality/kolibri-docs/pull/52#pullrequestreview-208671830

benjaoming:
Yes, we need to change this. The tutorial was written before `kolibri-server` was made.

radinamatic:
@benjaoming Could you take this on and update the RPi tutorial?

benjaoming:
It's been in my TODO for a few weeks :sob:

radinamatic:
Fixed in #118

PR #116, opened 2019-07-24
Update docs for 0.12.6/7 release

### LINKED ISSUE #107: update guidance on running kolibri.exe from command line
Pending https://github.com/learningequality/kolibri-installer-windows/issues/102

Once `kolibri.exe` is available from the system path, we can simplify the instructions, removing references to `C:/PythonXY/Scripts`

### LINKED ISSUE #115: hardware requirements should be distinct for Windows and Linux
Current [server requirements](https://kolibri.readthedocs.io/en/latest/install.html#servers):

> image

@jredrejo notes that:

> I think we should have different requirements on Windows and Linux. On Linux you can run it with 1 GB of RAM but I doubt you can even boot a Windows 10 with just that amount of memory.

radinamatic:
Ready for review!

- [X] Renamed and moved the section about command line management among the first level TOC items
- [X] Fixes #115 (Augment Windows RAM reqs)
- [X] Fixes #107 (Remove the path requirement on Windows, as `kolibri` is now accessible in command prompt from anywhere)
- [X] Updated languages
- [X] Added pending suggestions from #106
- [X] Added note about the port env var taking precedence over ``option.ini``
- [X] Reworded the ``kolibri-server`` section as **Higher Performance with...**

Selection_205

- [X] Added the ``Hard drive provisioning of multiple Kolibri servers`` page, which originated as a Google doc to support other partners and grantees. I totally see this as a _tutorial_, so given that they are now **2**, I've put them back in the TOC!  

Selection_206


docs/install/_linux.rst
```diff
@@ -136,6 +109,35 @@ To change the system service owner, you need to change the configuration of the
 .. note:: Replace the ``$USER`` in commands above with the name of the user you wish to be the new Kolibri system service owner.
 
 
+Optimized Performance with ``kolibri-server`` package
+-----------------------------------------------------
+
+We finalized the release of an optimized version of Kolibri for Linux systems called ``kolibri-server``. This version is recommended for:
+
+* Clear server-client setup for implementations with many client devices
```
indirectlylit: > Clear server-client setup

Not sure what this means

docs/install/_linux.rst
```diff
@@ -136,6 +109,35 @@ To change the system service owner, you need to change the configuration of the
 .. note:: Replace the ``$USER`` in commands above with the name of the user you wish to be the new Kolibri system service owner.
 
 
+Optimized Performance with ``kolibri-server`` package
```
indirectlylit: "Optimized" isn't quite the right word here.

Maybe "Higher Performance with the `kolibri-server` package"

docs/install/_linux.rst
```diff
@@ -136,6 +109,35 @@ To change the system service owner, you need to change the configuration of the
...
+-----------------------------------------------------
+
+We finalized the release of an optimized version of Kolibri for Linux systems called ``kolibri-server``. This version is recommended for:
```
indirectlylit: maybe:

> We have of a higher-performance wrapper around the `kolibri` package for Ubuntu and Debian systems called `kolibri-server`. This version is recommended for:

docs/install/_linux.rst
radinamatic: I believe it was supposed to differentiate from single device installation, that might not have clients to serve.

docs/install/_linux.rst
radinamatic: Thanks, amended!

docs/install/_linux.rst
radinamatic: Done!

docs/install/_linux.rst
indirectlylit: specifically I don't know what the word "Clear" means in this context

docs/install/_linux.rst
```diff
@@ -136,6 +109,35 @@ To change the system service owner, you need to change the configuration of the
...
 
+Higher Performance with the ``kolibri-server`` package
+------------------------------------------------------
+
+We have of a higher-performance wrapper around the ``kolibri`` package for Ubuntu and Debian systems called ``kolibri-server``. This version is recommended for:
+
+* Clear server-client setup for implementations with many client devices
+* When the server device has more than one CPU core
+
+``kolibri-server`` has some specific features like **static content cache** and **optimized usage of the CPU processing power** in multi core devices. Typical use case would be a *Raspberry Pi 3+* device with 4 CPU cores, which will perform much better and faster with ``kolibri-server`` compared to base Kolibri.
```
benjaoming: I think we should be bold enough to be very clear about the real benefits of using `kolibri-server`:

```suggestion
``kolibri-server`` has some specific features like **static content cache** and **optimized usage of the CPU processing power** in multi core devices. For example, a *Raspberry Pi 3+* device has 4 CPU cores, and with ``kolibri-server`` it is likely to serve 2-3x as many clients.
```

docs/install/_linux.rst
benjaoming: I think you can remove the text about what it's recommended for. That will leave the information from the first paragraph and the one preceding the list: Any system with Debian/Ubuntu, gives 2-3x the performance on a Raspberry Pi. I used "2-3x" in the example to be a bit conservative. But I believe @jredrejo has indicated that it might be more like 3-4x from benchmarks.

docs/install/tutorials/provision.rst
```diff
@@ -0,0 +1,148 @@
+:orphan:
+.. _provision:
+
+Hard drive provisioning of multiple Kolibri servers
+===================================================
```
benjaoming, line 5: Have you seen this already? https://kolibri.readthedocs.io/en/latest/install/tutorials/rpi.html#saving-your-image-for-replication

Not sure if useful for this guide, but I added `provisiondevice` in that guide.

I guess the Raspberry Pi tutorial could perhaps cross-reference to this one? I.e. "For more general advice, read ..."

Since this is also a tutorial, should we kick off the Tutorial section? :)

"Provision" is not a normal word in my vocabulary, so would also perhaps suggest to find an alternative for the headline. "Copying a Kolibri setup for re-use" or similar.

docs/advanced.rst
```diff
@@ -217,49 +218,49 @@ Change Language
...
-| Burmese               | ``my``          | 
-+-----------------------+-----------------+
++--------------------------+-----------------+ 
+| Language                 | <langcode>      |
++==========================+=================+ 
+| English                  | ``en``          |
++--------------------------+-----------------+
+| Spanish (Spain)          | ``es-es``       | 
++--------------------------+-----------------+ 
+| Spanish (Latin America)  | ``es-419``      |
```
benjaoming, line 228: Haven't followed this change -- will it break existing `es-mx` setups, since they may have a non-existing language code default?

benjaoming (commented):
Gave it the time that I had at my hands..

benjaoming:
Great changes btw! :smiley:

docs/install/_linux.rst
radinamatic: Removed!

docs/install/_linux.rst
radinamatic: Added your suggestion! :slightly_smiling_face:

docs/install/tutorials/provision.rst
radinamatic: Cross-referenced to RPi tutorial, and put back `Tutorials` in the TOC :tada:

docs/advanced.rst
radinamatic: We never had `es-mx` in Kolibri, so this is not break anything. We had it for a short time on Studio, but it was never maintained properly. Since it's not sustainable to maintain several Spanish country-locales, we introduced `es-la` (or `es-419` that is used for Latin American Spanish) on Kolibri, and will have one generic `es` on Studio.

radinamatic:
@indirectlylit Would you give it one last look? Thanks!

docs/install/tutorials/provision.rst
```diff
@@ -13,7 +13,10 @@ You will need an **active Internet connection only to preload content channels o
 Required hardware
 -----------------
 
-#. Laptops and/or other devices to host Kolibri servers.
+#. Laptops and/or other devices to host Kolibri servers. 
+
+	.. tip:: If you are planning to use Raspberry Pi as your server device, read our comprehensive tutorial about setting up :ref:`tutorial_rpi`.
```
indirectlylit: we might want to wait to advertise this until #104 is addressed

benjaoming: `kolibri-server` isn't a seemless addition to the tutorial, that is true.

radinamatic: But RPi can certainly have the vanilla Kolibri running on it, it just won't be able to make the best use of its own resources, correct?

In other words, if the RPi tutorial is _still_ valid as such (but we should update it as soon as possible), I don't see why would that be a blocker to this PR... :confused:

benjaoming: My concern would be the level of advertisement: It's hard to move people from the unchanged RPI tutorial to a new one because the current one needs some adaptation to include `kolibri-server`.

How about this: I fix this during this week, and then we don't worry about the level of advertisement?

radinamatic: Ok, valid concern. :+1: 

@indirectlylit should we wait for @benjaoming to fix the RPi tutorial, and then release the updates for the docs, or it's better to hold on (revert) the tutorial related changes and release the docs without them until the fix is in?

indirectlylit: I think if you just comment out this one line, we can merge the changes

radinamatic: Done and ready @indirectlylit! 🙂

PR #125, opened 2019-08-21
Delete 'IP' 

radinamatic:
TIL that _IP address_ with specified port is not, in fact, **IP** 🙂


indirectlylit (dismissed):

indirectlylit:
actually, I think this should target `0.12.x`, not `develop`

radinamatic:
🤦‍♀ 

Fixed!

PR #118, opened 2019-08-01
RPi tutorial update

benjaoming:
This is ready for review -- or: I won't have time to improve this immediately. But it's one of the steps for kolibri-server to be added to the stable PPA.

This cannot be published before those binaries are copied in though..

Closes #104

PR #127, opened 2019-08-27
Fix the path for the access point configuration file

jredrejo:
The file for the hostapd configuration is `/etc/hostapd/hostapd.conf` , not `/etc/hostapd/hostapd.conf`


benjaoming:
This was fixed here: https://github.com/learningequality/kolibri-docs/pull/118/files#diff-defb699037c96eefc6909e020c1b1034R185

Since @jamalex also contributed fixes that were already in said PR, I'm going to approve it myself very soon, before the current docs attract more double fixes :)

PR #128, opened 2019-09-04
Add instructions to prevent issues on NComputing systems

### LINKED ISSUE #126: add information about running Kolibri in an NComputing environment
we've had a couple instances now of people using [NComputing setups](https://www.ncomputing.com/products/vSpace/vSpace%20PRO) and running into problems.

ref: https://learningequality.slack.com/archives/C0LKF6TRD/p1566499868052700

from @jamalex 

> if you set the Kolibri windows installer to autostart Kolibri at boot, it does so for every account. Every terminal on an n-computing cluster is a new account --  it's all the same computer (one IP -- and same 127.0.0.1), but different home directories depending on who logged in first.
> 
> So it should be disabled from autostarting on every account except the main admin account, and then the teacher needs to log in using that account after the computer is turned on to start kolibri up
> 
> Installing Kolibri also put the Kolibri launcher icon on the desktop of every student account so we removed those and instead added a URL shortcut icon

We should update the docs to give people this info

radinamatic:
Now it's ready for review!

Fixes #126

### DIFF
--- a/docs/access.rst
+++ b/docs/access.rst
@@ -31,6 +31,19 @@ While it is running, **Kolibri** will display an icon in the Windows taskbar (us
 .. tip:: 
   If you close the browser window, Kolibri will still be running in the background. Use the **Load in browser** menu item to reopen it in the browser.
 
+
+.. _ncomputing 
+
+.. warning::
+  If you install Kolibri in settings with various virtual desktops running on the same single hardware device (for example with NComputing), you need to make sure that Kolibri is running **only from the admin account (terminal)**. Confirm that the **Run Kolibri at system startup** option is checked when the admin logs in, and unchecked with all the other user accounts.
+
+  It is also recommended that you remove the Kolibri desktop laucher icon from all non admin accounts in this type of setup, and substitute it with a regular shortcut that opens the Kolibri server IP. To add the desktop shortcut follow these steps.
+
+  #. Right click anywhere on the desktop.
+  #. Select *New > Shortcut*.
+  #. Type in the Kolibri server IP (most probably ``http://127.0.0.1``), save and exit.
+  #. Double-click the shortcut to test if it opens the browser and loads Kolibri.
+
 .. note::
   Remember to :ref:`configure other computers <access_LAN>` in the network to access **Kolibri** content.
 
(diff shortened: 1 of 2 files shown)


docs/access.rst
```diff
@@ -31,6 +31,19 @@ While it is running, **Kolibri** will display an icon in the Windows taskbar (us
 .. tip:: 
   If you close the browser window, Kolibri will still be running in the background. Use the **Load in browser** menu item to reopen it in the browser.
 
+
+.. _ncomputing: 
+
+.. warning::
+  If you install Kolibri in settings with various virtual desktops running on the same single hardware device (for example with NComputing), you need to make sure that Kolibri is running **only from the admin account**. Confirm that the **Run Kolibri at system startup** taskbar option is checked only when the admin logs in, but unchecked when any other user of virtual desktops is logged in.
```
benjaoming:
```suggestion
  In some Windows multi-user environments (for example NComputing), you need to make sure that Kolibri is running **only from the admin account** to avoid having Kolibri started for each login/session. Confirm that the **Run Kolibri at system startup** taskbar option is checked only when the admin logs in, but unchecked when any other user of virtual desktops is logged in.
```

benjaoming (changes requested):
Really great thorough work! Wondering if this problem pertains a wider range of Windows systems. To be sure, tried to adjust the first sentence of the warning box to be more clear on who it could affect.

benjaoming:
Reflection: The warning box is quite long, and I know it's hard to do anything about. Had the support issue been dealt with in the Community Forums, we would have been able to put together documentation that referenced more instructions in a forum thread, which might in some cases be better for concise documentation.

I'm wondering if questions that have been raised outside of the normal community forum channels can still be added there :)

radinamatic:
> The warning box is quite long, and I know it's hard to do anything about.

Yep, I wanted it to stand out hence the warning box in the Windows access section, but this type of info would really belong in a _knowledge base_ type of document, that we do not have at the moment. 

The "question" did not come from the community forums, but from our field experience in Uganda. However, I did add a reference for it, so it can be easily linked to in the forums or in email communication, for example. Once this PR is merged, [this link](https://kolibri.readthedocs.io/en/latest/access.html#ncomputing) will bring the user directly to the warning.

PR #131, opened 2019-09-07
Update tutorial to make it more clear that we also run on RPi 4

benjaoming:


docs/install/_linux.rst
```diff
@@ -138,6 +138,4 @@ To install ``kolibri-server`` follow these steps.
 Raspberry Pi
 ------------
 
-To install Kolibri on RPi refer to our :ref:`Raspberry Pi Tutorial for Kolibri <tutorial_rpi>`.
-
-.. warning:: Kolibri is intended for **Raspberry Pi Model 3** and upwards.
+To install Kolibri on RPi refer to our :ref:`Raspberry Pi Tutorial for Kolibri <tutorial_rpi>`. Kolibri is tested and designed to work on **Raspberry Pi Model 3 and 4**.
```
jredrejo: Model 3, 3+ and 4 or just 3+ and 4?

Why are we not not including Model 2?  because it's not tested or because it's known it does not work?
Also I think @jamalex  has tested it succesfully in Pi Zero W.

jredrejo (changes requested):

docs/install/_linux.rst
benjaoming: Sure, we can add other models that we tested on.

Model 2 needs an external Wi-Fi adapter, which the tutorial doesn't cover. I tried the latest Raspbian Buster and I don't think it works on Model 3 - which is fine, since the guide essentially covers Stretch and Buster. But that makes me wonder if Model 2 supports Stretch, because I don't recall.

jredrejo: According to https://en.wikipedia.org/wiki/Raspbian , Model 3 works with Buster, and Model Pi Zero W has been tested by @jamalex so I think we should include 3, 3+ , 4 and Zero W.
For Model 2, I agree. If the tutorial does not cover installing an external usb dongle we should not mention it. It should work with Buster and we might mention it, but not as part of this tutorial.

benjaoming: Adapted the instructions in the install and tutorial to express a less rigid set of Raspberry Pi models.

radinamatic (approved):
Looks good, merge away! 👍

PR #132, opened 2019-09-12
Warning to include "http://" when accessing the server

### LINKED ISSUE #129: Ensuring that users add "http://" when accessing the server
It's not always clear to users in the documentation that "http://" must be included in the URL when accessing Kolibri via server. https://kolibri.readthedocs.io/en/latest/access.html#access-kolibri 

We've also seen before where people assume it's "https" so this could be helpful for addressing that as well. 

Suggestion to add a "tip" to the documentation.

radinamatic:
Fixes #129.

### DIFF
--- a/docs/access.rst
+++ b/docs/access.rst
@@ -119,7 +119,7 @@ For example, if Kolibri is installed and started on a computer with the address 
 
 
 .. warning::
-  * When limited to using older browsers it may be necessary to **explicitly include** the ``http://`` in front of the IP in the address bar, for example ``http://192.168.8.134:8080``. Many modern browsers may also take the IP address or the local network hostname as a search term and redirect you to Google. In other cases, browsers may try to use the secure **HTTPS** protocol by default and add an **S**  automatically (``httpS://``), which may also cause issues while running Kolibri.
+  * When entering the URL in the browser, it may be required to explicitly include `http://`, for example `http://1.2.3.4:8080` or `http://1.2.3.4:80`. Many browsers will incorrectly interpret an IP address or local network hostname as a search term. In other cases, browsers may incorrectly add `https://` instead of `http://`.
   
   * In case you decide to make Kolibri available on the port 80, instead of the default 8080, you **must always include** ``http://`` in front of the server's IP.
     
--- a/docs/access.rst
+++ b/docs/access.rst
@@ -119,10 +119,10 @@ For example, if Kolibri is installed and started on a computer with the address 
 
 
 .. warning::
-  * On older operating systems and browsers it might be necessary to **explicitly include** the ``http://`` in the address bar when accessing the server's IP. Recent browser versions should redirect it properly even if you write just the IP, but be sure to include ``http://`` if you have Kolibri running on older devices and encountering issues accessing the server's IP.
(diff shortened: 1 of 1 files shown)


benjaoming (changes requested):
I'm thinking the text could explain this a bit differently, with a full background of the issue:

When typing in your server's address in the browser address bar, take care to add `http://` in the first part of the address, for instance `http://1.2.3.4`. Many modern browsers may understand an IP address or local network hostname as a search term and redirect a user to Google. In other cases, browsers may add `httpS://` automatically, which will also cause issues.

radinamatic:
Thanks for the suggestion @benjaoming, I rephrased to adjust! 👍

docs/access.rst
```diff
@@ -118,6 +118,12 @@ For example, if Kolibri is installed and started on a computer with the address
   * You can also use the ``ipconfig`` command on Windows or ``ifconfig`` command on Linux/OSX to find the externally visible IP address of the device running the Kolibri.
 
 
+.. warning::
+  * When limited to using older browsers it may be necessary to **explicitly include** the ``http://`` in front of the IP in the address bar, for example ``http://192.168.8.134:8080``. Many modern browsers may also take the IP address or the local network hostname as a search term and redirect you to Google. In other cases, browsers may try to use the secure **HTTPS** protocol by default and add an **S**  automatically (``httpS://``), which may also cause issues while running Kolibri.
```
jredrejo: I don't thin `http://192.168.0.134:8080` is a good example. The browser will redirect to https when the port is not used. I.E: `http://192.168.0.134`will be redirect to `https://192.168.0.134` because the port is not added. If the port is added it keeps `http`.
That's why I mentioned the problem exists *only* when using port *80*, because this was the former implicit port and now it has been changed to *443* (https) by the browsers since they implemented https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security

jredrejo (changes requested):

docs/access.rst
benjaoming:
```suggestion
  * When typing in your server's address in the browser address bar, it may be required to add `http://` in the first part of the address, for instance `http://1.2.3.4:8080` or `http://1.2.3.4:80`. Many modern browsers may understand an IP address or local network hostname as a search term and redirect a user to Google. In other cases, browsers may add `httpS://` automatically, which will also cause issues.
```

@radinamatic notice that the original issue pertained some kind of "older browsers" and @jredrejo is referring to current and upcoming browsers. I think it's best to just describe the behavior and add a simple-to-read caution.

If people are setting up Kolibri on Port 80 and configuring their own host names, i.e. `http://kolibri`, they will probably understand the caution, given that they have a bit of technical background required to make a choice of port 80 :)

benjaoming (changes requested):
The issue can affect both old and new browsers.

docs/access.rst
benjaoming: I'm not sure how browsers will implement this in the future, but I think the warning is a reasonable caution such that everyone can go ahead and make their own observations / troubleshooting with some pretty solid hints.

benjaoming (changes requested):
@radinamatic I made a suggestion here. Once it's in, this PR LGTM :+1:

docs/access.rst
indirectlylit: We can address Jose's feedback by adapting the text slightly

```suggestion
  * When accessing Kolibri by IP, in many browsers it is necessary to **explicitly include** the ``http://`` in front of the IP in the address bar. For example, this might be ``http://192.168.8.134:8080`` for port ``8080`` or ``http://192.168.8.134`` for port ``80``. Omitting ``http://`` may result in undesirable redirects.
```

benjaoming: I already made a suggestion for this, see below ;)

indirectlylit:
updated PR based on feedback from @benjaoming and @jredrejo

PR #139, opened 2019-11-29
Fixes wrong paths in fixing malformed db issue

benjaoming:
Noting that there were errors in the previous instructions.


jredrejo (approved):
Nice changes!

Issue #45, opened 2018-09-01
Header anchor IDs should be semantic and suitable for permalinks


benjaoming:
For instance: `https://kolibri.readthedocs.io/en/latest/install.html#id8`

In this example, `#id8` is enumerated, and adding another heading before will change it into `#id9`, thus breaking outside references.

If it's offered by Sphinx, we should find a way to change these links so that they are permanent. Otherwise a lot of references from Community support etc. will be broken.

If this doesn't work, a nice alternative/compromise would be that we simply don't link to specific sections with anchor links that are non-permanent.

indirectlylit:
sorry, could you clarify what the expected behavior is?

benjaoming:
Yup @indirectlylit, updated the description

indirectlylit:
Got it - thanks. Agreed, the ID should be derived from the header text

radinamatic:
I believe it's happening because the page contains multiple 3rd level headings with the same name: Compatibility, Install, Uninstall, Upgrade. When a heading/subheading cannot be uniquely resolved as `page.html#heading`, it defaults to resolving as a numbered `id` one.

I've been researching solutions, but there does not seem to be a suitable one. I could try to create a more explicit header links, but they will not be included in the TOC ([example](https://stackoverflow.com/questions/23530950/how-to-set-custom-headerlink)). Not many solutions offered for customizing the [TOC directive](http://www.sphinx-doc.org/en/stable/markup/toctree.html) with Sphinx. 

Another option would be to make headings unique: _Compatibility on Windows, Install on Windows, Uninstall on Windows, Upgrade on Windows_ but I don't like the sound of it, it's repetitive... 😕

benjaoming:
@radinamatic thanks for investigating - yup, let's not compromise the wording of the headers themselves :)

radinamatic:
Actually, I would prefer to keep this open. 

I've been working, on and off, to find a solution for this, but it is a major undertaking as it would imply the restructure of the way topics are presented: instead of using the `.. include` directive, to form one long page for the first level TOC items, subtopics would be loaded as single pages and navigation would mean more clicking to <kbd>Next</kbd> and <kbd>Previous</kbd> buttons, instead of scrolling the page up and down.

These shorter more granular pages would ideally allow fully semantic permalinks, even with multiple similar headings, for example: 
- `https://kolibri.readthedocs.io/en/latest/install/linux.html#upgrade`
[...]

PR #142, opened 2019-12-13
restructure install section and clarify raspberry pi install options

indirectlylit:
* refactors 'install' section according to https://github.com/learningequality/kolibri-docs/issues/45 and https://github.com/learningequality/learningequality.org/pull/479
* adds stub section for mac
* adds initial language clarifying the relationship between the raspberry pi OS image and the tutorial
* delete dead/draft/hidden content


indirectlylit:
Note to all: if this looks like a reasonable base to work off of, please merge in and submit additional PRs to that so I don't become a bottleneck

benjaoming (approved):

benjaoming:
It's about going through steps and learning them, so I think it's a good idea to retain the "tutorial" as a deliberate part of the format, visible to both documentation maintainers and readers.

docs/install/index.rst
```diff
@@ -0,0 +1,38 @@
...
+See the `Kolibri changelog here <https://learningequality.org/r/kolibri-changes>`_ to review specific changes in each version of Kolibri. Documentation for `previous Kolibri versions are available here <https://readthedocs.org/projects/kolibri/>`_.
+
+
+Learn about whether Kolibri will run on your system, and how to get the server set up:
+
+.. toctree::
+  :maxdepth: 1
+
+  system_requirements
+  initial_setup
```
radinamatic, line 15: This step (our _setup wizard_) logically goes **after** the installation, so I don't think this is the best place.

radinamatic:
I rather liked *Tutorials* being a separate section, so not sure why the decision to eliminate them...? Maintaining that section would not be in conflict with the rest of your proposed additions here.

Current | Proposed
------------ | -------------
 | 

I will approve and merge, but will move **Initial setup** back at the end in the upcoming update, and would like to reinstate tutorials as a separate sections.

indirectlylit:
My motivation for removing the 'tutorials' section was following the guidance in https://github.com/learningequality/kolibri-docs/issues/45 and similar work in the developer docs here: https://github.com/learningequality/kolibri/pull/4536 - namely, that flat is better than nested.

It allowed the top-level "Install" page to be formatted thusly:

image

However, if you guys prefer a sub-section for tutorials that's fine with me.

Regarding the re-ordering of the items, I put then in order of generality and relevance, rather than sequence of use. In other words, everyone needs the system requirements and setup wizard docs, but the others are more context specific.

However, again I defer to you on this.

radinamatic:
I understand what you are saying. 

I like the top-level "Install" page formatting, and will try to implement that for the rest of pages according to what we discussed in #45 (although that might be a bigger refactor, to be polished in the subsequent updates after the 0.13 release), and I'll see if it's doable to make it into this update.

radinamatic (approved):
Thank you! 👍

benjaoming:
Agreeing about flat structure... The number of clicks required to reach something is great to keep down, having good and well-placed cross references is important.

I was concerned about eliminating "Tutorial" as a word and sticking to the format. Because install instructions, to me, are something different and that an "instruction" framing will erode the tutorial-oriented text over time.

Installation instructions = as simple and minimal as possible
Tutorials = with learning purposes

The other thing about tutorials is that they are written with their own scoped specifications: "In this tutorial, we show you how you can do X on platform Y".

I like tutorials, I think we should have more :)

Issue #144, opened 2019-12-15
consider recommending --user flag for pip install


indirectlylit:
currently for pip we simply say

```bash
pip install kolibri
```

In fact, more likely the user would want to do

```bash
pip install --user kolibri
```

or, in slightly more rare scenarios

```bash
sudo pip install kolibri
```

PR #147, opened 2019-12-19
Add better guidance for installing Kolibri from pip

benjaoming:
@indirectlylit commented on this somewhere and I wrote a PR instead of commenting back that I agree :)

I also added a not that this makes the `kolibri` command available from command line, since this is slightly different from other installers that add system services and short cuts.

I think that users of virtualenvs will know very well not to use `--user` or `sudo`, so an instruction for them should be superfluous.

### DIFF
--- a/docs/install/python.rst
+++ b/docs/install/python.rst
@@ -12,7 +12,13 @@ You can install Kolibri as a standard package from PyPi (works on Mac, Windows, 
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


indirectlylit:
> @indirectlylit commented on this somewhere

ref: https://github.com/learningequality/kolibri-docs/issues/144

indirectlylit (approved):
thank you

PR #156, opened 2020-02-03
Add French translation

radinamatic:
PO files for the French version of the user docs, translated & proofread.


benjaoming:
Reviewed: Going into the correct branch and after quickly studying French, I was also able to verify that the translations are correct :D

Issue #48, opened 2018-09-11
clarify and improve availability of older installer versions


indirectlylit:
Migrated from https://github.com/learningequality/kolibri/issues/4208

-----

Current state:

* In http://kolibri.readthedocs.io/, we intentionally allow people to view docs for older versions
* If someone intentionally navigates to an older version and tries to download an installer, they are redirected to the most recent version
  * On one hand, this is good because it encourages users to upgrade
  * On the other hand, this is bad because it's misleading, since the user may be intentionally trying to download an older version (e.g. to see if an issue they're having is a regression, or to re-install in a broader implementation that is still using a legacy version)

Needs:

1. Users should be encouraged to upgrade to the latest version whenever they are viewing docs or downloads for older versions, in case they got there unintentionally
2. Users should have an officially-supported and discoverable way to download any previous release, if that's what they really want
3. Downloads should have appropriate analytics tracking (e.g. we get better information from website redirects than through pantry nginx logs)

I don't think this is an urgent problem to solve, but in the long run I think we'll likely want to make a combination of changes to pantry, the website, and the docs to satisfy the three needs above.

benjaoming:
I agree. It cannot be a question of *if* users can land on out-dated documentation because of search engines or links from other sites.. it must be a question of *how many*.

Suggested fix:

1. Do not refer directly to the latest installer. Instead, refer to the download page. New installer types appear.
2. Display the Kolibri version on all the installation instruction pages.

All fixes would benefit from being back-ported to older release branches.

A small note: If the user is displayed the version of Kolibri which the docs were written for, that may still not lead them to conclude "this is out-dated" since she may not be aware of the latest version. However, always referring people to the main download page should...

benjaoming:
Suggesting that we use what RTD offers:

![image](https://user-images.githubusercontent.com/374612/47256331-b3a98800-d47e-11e8-9d94-fd1fc4c31d52.png)

The requirement is that we start tagging the documentation using git tags. ReadTheDocs will automatically add the notification for people browsing old tags.

The disadvantage is that Kolibri is called a "library" in the notification, but we could probably live with this and just try to get that fixed upstream, suggesting to call it something more generic like "software" instead of "library".

benjaoming:
@radinamatic @indirectlylit any comments here?

You can refer to the Wagtail docs for a look at how this ends up... https://docs.wagtail.io/en/latest/

radinamatic:
> start tagging the documentation using git tags

[...]

PR #157, opened 2020-02-03
updates for 0.13.1 release

### LINKED ISSUE #148: document any best practices related to upgrading and trying new versions and installation methods
We've observed some issues related to trying new versions and installation methods. One example:  https://github.com/learningequality/kolibri/issues/6388

There are likely many other examples. Let's try brainstorm and document some best practices to avoid situations that we expect might cause issues.

### LINKED ISSUE #153: Troubleshooting: Malformed Database can also happen on other SQLite db files
Pending a possibly automatic resolution of "malformed database" occurring for `~/.kolibri/job_storage.sqlite3` and `~/.kolibri/process_cache/cache.db`, we should perhaps expand our Troubleshooting section here:

https://kolibri.readthedocs.io/en/latest/manage/get_support.html#troubleshoot-database-issues

The resolution for the other files is to simply delete them entirely and start Kolibri again, as they should be automatically created.

radinamatic:
- [X] Fixes #153 
- [X] Document `exportlogs` command
- [X] Document the [data syncing](https://github.com/learningequality/kolibri/pull/6028) feature 
- [X] Added `robots.txt` file to [prevent older versions of the docs being indexed](https://github.com/learningequality/kolibri-docs/issues/48#issuecomment-581177092) by search engines 
- [X] Added the warning that unlisted channel must be published before using the token ([comment](https://learningequality.slack.com/archives/C4HNGQJGH/p1580223390012700) from Slack)
- [X] Added the recommendation to use Chrome instead of open source browsers in case of the video codec issues ([comment](https://learningequality.slack.com/archives/C4HNGQJGH/p1580504644015500) from Slack)
- [X] Added warning to stop Kolibri before upgrading. Fixes #148


docs/conf.py
```diff
@@ -27,6 +27,7 @@
 extensions = [
     'sphinx.ext.todo',
     'sphinx.ext.autodoc',
+    'sphinx_rtd_theme',
```
benjaoming, line 30: Oh interesting that this was missing, did it cause any issues?

docs/manage/get_support.rst
```diff
@@ -36,22 +36,33 @@ In case you receive the ``database disk image is malformed`` error in Terminal,
...
+      sqlite3 db.sqlite3 .dump | sqlite3 fixed.db
+      mv fixed.db db.sqlite3
+
+#. Remove the following files.
+   
+    .. code-block:: bash
+
+      rm -f db.sqlite3-wal db.sqlite3-shm job_storage.sqlite3 /process_cache/cache.db
+
+#. Restart Kolibri.
```
benjaoming, line 65: Just want to make sure that this is intentional: The steps above are no longer in a copy-paste friendly block. That's okay, copy-pasting long commands is a bit of a mal-practice, although I think these lines are all very short and readable.

But by not being in the same block, it's now implicit that everything follows `cd ~/.kolibri`, and cannot be run individually.

As Kolibri is likely crashed during this operation, we didn't put "start Kolibri" in the beginning, and finished it with " After this, you should be able to start Kolibri again." - restart makes slightly less sense when it's not running.

docs/conf.py
radinamatic: No issues, but if I understood correctly the [theme docs](https://sphinx-rtd-theme.readthedocs.io/en/latest/installing.html), the line is needed so the localized theme strings can be displayed correctly. We still don't have localized versions of the 0.13 docs, but we'll hopefully have French in a couple of months.

docs/manage/get_support.rst
radinamatic: The reason why I decided to split the block, is because while working on supporting the translation to French, I realized that none of the in-block-comments were being extracted as translatable strings, and would remain in English. I think we should aim to have the code comments translated too, and write them in the way that permits string extraction.

Since the commands are in an ordered list, my assumption was that it was clear enough the commands needed to be run in a sequence, and not individually. 

Do you recommend putting the explicit `stop` (at the beginning) and `start` commands at the end of the list?

docs/conf.py
```diff
@@ -14,7 +14,7 @@
 # FOR! DO NOT LET THEM BE TARGETTED AT MORE THAN ONE MINOR SERIES!
 # I.E.: 0.1.x -- important to add 'dev' suffix for docs targetting development
 # series.
-DISPLAY_VERSION = "0.13.0"
+DISPLAY_VERSION = "0.13.1"
```
indirectlylit: I believe the outcome  of discussion in https://github.com/learningequality/kolibri-docs/issues/48 was that we would make this `0.13` and leave off the patch release?

radinamatic: My bad, I wasn't following that discussion :woman_facepalming:

docs/manage/get_support.rst
benjaoming: Ooooh that's why! I totally didn't realize that these code blocks could behave like this. Can whole blocks  be marked for translation?

radinamatic: Well, commands should remain marked as untranslatable (in between pairs of back ticks or behind ``.. code-block::``), so the only way I see is to separate comments and code lines completely. This will prevent c/p friendly blocks, but as you say, that is not kosher anyways  :slightly_smiling_face:

benjaoming: Thanks, I get what is at stake, let's not push code fragments together with translations, that seems wrong. The "hack" here was nice for non-translatable purposes otherwise, but not wort pursuing.

docs/conf.py
benjaoming: Thanks for the explanation :)

docs/conf.py
```diff
@@ -120,6 +121,12 @@
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
+
+# Add any extra paths that contain custom files (such as robots.txt or
+# .htaccess) here, relative to this directory. These files are copied
+# directly to the root of the documentation.
+html_extra_path = ['']
```
benjaoming: Can you tell me what you are trying to achieve here? The above reveals the whole source tree, including `conf.py` in the published HTML tree, which for various reasons is wrong.

image

docs/install/ubuntu-debian.rst
```diff
@@ -39,6 +39,8 @@ From the command line: ``sudo apt-get remove kolibri``.
 Upgrade
 *******
 
+.. warning:: Make sure that Kolibri is not running before upgrading.
```
benjaoming: This warning is unnecessary. The .deb will restarts Kolibri as part of upgrading. If it doesn't work, it's a bug... but in no way a documentation thing.

We cannot IMO in any consistent way require the user to switch off Kolibri before upgrading.

docs/install/windows.rst
```diff
@@ -42,6 +42,8 @@ Uninstall
 Upgrade
 -------
 
+.. warning:: Make sure that Kolibri is not running before upgrading.
+
```
benjaoming, line 46: If this is a Windows issue, I think we should refer to an issue report or Community Forum post where the reasoning for this warning is elaborated. Or just elaborate it here.

docs/manage/command_line.rst
```diff
@@ -228,6 +228,17 @@ If you need to permanently delete a Kolibri user and all the data associated wit
 .. warning:: This will permanently erase all the user data.
 
 
+Export Data Logs
+****************
+
+Kolibri database may become large after prolonged use, which can cause the data logs export process from **Facility > Data** the to be slow. Use the following command to export logs from the Terminal or command prompt.
```
benjaoming:
```suggestion
During prolonged usage, Kolibri's database increases in size, which slows down the data logs export process from **Facility > Data**. Instead of using the export in the user interface, you can use the following command-line interface to export logs: 
```

docs/manage/get_support.rst
```diff
@@ -36,22 +36,33 @@ In case you receive the ``database disk image is malformed`` error in Terminal,
...
-	cp -b db.sqlite3* malformed/
-        # Create a new database
-	sqlite3 db.sqlite3 .dump | sqlite3 fixed.db
-        # Move to the active location
-	mv fixed.db db.sqlite3
-        # Remove so-called "write-ahead log"
-	rm -f db.sqlite3-wal db.sqlite3-shm
-
-After this, you should be able to start Kolibri again.
+#. Open the Kolibri home folder in the Terminal.
```
benjaoming:
```suggestion
#. Open a terminal and change the current working directory to your Kolibri's home folder. Keep the terminal open and continue with the next commands.
```

docs/robots.txt
```diff
@@ -0,0 +1,10 @@
+User-agent: *
+Disallow: /en/release-v0.4.x/
+Disallow: /en/release-v0.5.x/
+Disallow: /en/release-v0.6.x/
+Disallow: /en/release-v0.7.x/
+Disallow: /en/release-v0.8.x/
+Disallow: /en/release-v0.9.x/
+Disallow: /en/release-v0.10.x/
+Disallow: /en/release-v0.11.x/
+Disallow: /en/release-v0.12.x/
```
benjaoming: robots.txt supports "glob" expressions, I guess this is an okay reference: https://developers.google.com/search/reference/robots_txt

So all of the above can be replaced with `Disallow: /*/release-v*/` - although I can't say how precise the expression is, it relies a bit on whether `*` matches `/` or not.

Alternatively, we can add just one `Allow: /*/latest/` and exclude everything else.

benjaoming (changes requested):

docs/install/ubuntu-debian.rst
indirectlylit: > If it doesn't work, it's a bug... but in no way a documentation thing.

Can't it be both?

We've seen many bugs lately related to corrupted DBs and simultaneous access. Until we're confident those are all addressed, this seems like reasonable advice.

Perhaps:

```
.. warning:: We recommend making sure Kolibri is not running before upgrading.
```

docs/manage/command_line.rst
indirectlylit: Do we have an open issue about speed differences between using the UI and the command line? I was not aware of this

docs/install/ubuntu-debian.rst
benjaoming: I don't think such a warning looks very good with no further reasoning?

CC: @jredrejo with regards to the way that Kolibri is stopped and changes to the systemd service, timeouts etc... is there a chance that Kolibri will not be properly stopped?

Call is here:
https://github.com/learningequality/kolibri-installer-debian/blob/master/debian/kolibri.scripts-common#L390

Implementation is here, noting that it will ignore if stopping the system service fails (presumably to ignore if it wasn't running in the first place)
https://github.com/learningequality/kolibri-installer-debian/blob/master/debian/kolibri.scripts-common#L154

docs/robots.txt
benjaoming: Note to maybe add the sitemap 
https://kolibri.readthedocs.io/sitemap.xml

Example: https://docs.readthedocs.io/robots.txt

docs/manage/command_line.rst
benjaoming: Me neither - @radinamatic ?

docs/conf.py
radinamatic: Corrected, thank you!

docs/install/ubuntu-debian.rst
radinamatic: As discussed, let's keep the warning but also open a high level issue that we may link in the warning, at least in the Debian section.

docs/install/windows.rst
radinamatic: Warnings like this are not at all unusual on Windows, I often restart the system before installing or upgrading anything, as a preventive measure :wink:

docs/manage/command_line.rst
radinamatic: The inspiration for documenting came from this [thread on the community forums](https://community.learningequality.org/t/0-13-generating-data-on-session-logs-takes-so-long-doesnt-seem-to-finish/1387).

benjaoming: Is the fundamental issue here is that this type of job lacks user feedback and a separate async job process?

Maybe it can be expressed differently:

> The Kolibri user interface currently does not display progress and estimated duration of data log exports. For larger databases, we recommend using the command line interface: (...)

indirectlylit: Based on the thread in the forums, I don't believe any of the suggestions are accurate.

I filed a new issue here: https://github.com/learningequality/kolibri/issues/6582

If think this is a common problem, we can make a note saying that

> If the log generation appears to hang, we recommend using the command line interface: ...

Otherwise I'd recommend removing this completely

radinamatic: I don't see the reason _not_ to document the command. I added the tip in the **Manage > Data** page that links to the command, where the inaccurate suggestions are removed leaving just the wording about what the user can observe (logs taking long time to generate).

benjaoming: This looks good and won't need an update once the issue filed by @indirectlylit is solved. Great work! :100:

benjaoming (approved):
LGTM! :rocket:

radinamatic:
Thank you! :pray:

PR #159, opened 2020-02-19
update mac installation instructions

indirectlylit:
The mac installation instructions are still a bit confusing and over-complicated. The nested list is also strangely formatted.

Trying to simplify things a bit here:

| before | after |
|--|--|
| image |image|

This also provides a direct link to the DMG due to https://github.com/learningequality/learningequality.org/pull/486

### DIFF
--- a/docs/install/macos.rst
+++ b/docs/install/macos.rst
@@ -11,6 +11,8 @@ Kolibri macOS app has been tested on High Sierra (10.13), Mojave (10.14), and Ca
 Install
 -------
 
+.. note:: The Kolibri Mac app is not signed, which means that you must tell MacOS to allow it to run. Read `more information about opening unsigned apps <https://support.apple.com/guide/mac-help/open-a-mac-app-from-an-unidentified-developer-mh40616/mac>`_ from Apple.
+
 #. Download the `.dmg installer <https://learningequality.org/download/>`_ for Kolibri **version 0.13**.
 #. Double-click the downloaded ``.dmg`` file.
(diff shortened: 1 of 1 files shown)


radinamatic:
Can we please first get the #157 merged?

benjaoming:
@radinamatic not sure about the timing of this, but we can link up if you would like some tricks on how to quickly merge @indirectlylit 's PR so it doesn't interfere with other work you're doing?

radinamatic:
I don't plan to add anything more in #157 (no tricks are needed), if there are no remaining blockers, let's merge it.

benjaoming:
Ah okay, I get you! It's merged!

radinamatic (approved):
Great editing work, looks much better, thank you!

docs/install/macos.rst
```diff
@@ -13,7 +13,7 @@ Install
 
 .. note:: The Kolibri Mac app is not signed, which means that you must tell MacOS to allow it to run. Read `more information about opening unsigned apps <https://support.apple.com/guide/mac-help/open-a-mac-app-from-an-unidentified-developer-mh40616/mac>`_ from Apple.
 
-#. Download the `.dmg installer <https://learningequality.org/download/>`_ for Kolibri **version 0.13**.
+#. Download the `.dmg installer <https://learningequality.org/r/kolibri-mac-latest>`_ for Kolibri **version 0.13**.
```
indirectlylit: updating to a direct link due to  https://github.com/learningequality/learningequality.org/pull/486

PR #164, opened 2020-04-02
Fix error in "malformed" restore steps - add tip for using .recover

### LINKED ISSUE #162: Malformed DB guidance
I was following the instructions provided here: https://kolibri.readthedocs.io/en/latest/manage/get_support.html#troubleshoot-database-issues

but wasn't successful as I was trying to fix a malformed db. Instead I got returned a fix db that was empty.

@rtibbles and @lyw07 helped me debug, ran this command instead:
`sqlite3 corrupted_db_name .recover | sqlite3 fixed.db`

`.recover` was introduced last year in version 3.29: https://sqlite.org/changes.html

Request: It would be good to add .recover to the docs, but with the version caveat.

benjaoming:
Fixes #162

@radinamatic if you have changes, can you use the Github Suggestion feature?

CC: @jeepurs

### DIFF
--- a/docs/manage/get_support.rst
+++ b/docs/manage/get_support.rst
@@ -29,12 +29,14 @@ Windows systems often come with firewalls bundled and enabled, and this may inte
 You can try temporarily disable your firewall to see if it helps with connecting to Kolibri. If so, you'll want to turn the firewall back on and then create a exception rule for Kolibri to allow access.
 
 
-Troubleshoot Database Issues
-----------------------------
+.. _malformed_db:
 
-In case you receive the ``database disk image is malformed`` error in Terminal, try running these commands.
+Malformed database
+------------------
 
-The ``sqlite3`` command is necessary. This can be installed with ``sudo apt install sqlite3`` on a Debian-based system.
+In case you receive the ``database disk image is malformed`` error in Terminal, it does not mean that the database cannot be fixed. The error occurs in cases where Kolibri has been shutdown uncleanly, for instance due to a power outage or a software exception.
+
+.. tip:: The ``sqlite3`` command is necessary. This can be installed with ``sudo apt install sqlite3`` on a Debian-based system.
 
 #. Open a terminal and change the current working directory to your Kolibri's home folder. Keep the terminal open and continue with the rest of the commands.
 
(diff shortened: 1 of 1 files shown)


docs/manage/get_support.rst
```diff
@@ -56,16 +56,38 @@ The ``sqlite3`` command is necessary. This can be installed with ``sudo apt inst
       sqlite3 db.sqlite3 .dump | sqlite3 fixed.db
       mv fixed.db db.sqlite3
 
-#. Remove the following files.
+#. Remove temporary database files.
    
     .. code-block:: bash
 
-      rm -f db.sqlite3-wal db.sqlite3-shm job_storage.sqlite3 /process_cache/cache.db
+      rm -f db.sqlite3-* job_storage.sqlite3* notifications.sqlite3* process_cache/cache.db
```
benjaoming, line 65: Several changes:

* Note the forward slash `/` removed prepending the `process_cache` got removed.
* I also added `notifications.sqlite3`
* All paths got a little asterisk `*` so to remove both `-wal` and `-shm` which weren't restricted to just  `db.sqlite3`

docs/manage/get_support.rst
```diff
@@ -56,16 +56,38 @@ The ``sqlite3`` command is necessary. This can be installed with ``sudo apt inst
...
 
 For further assistance, please report the issue on our `Community Forums <https://community.learningequality.org/>`_, stating the operating system and Kolibri version.
 
+.. tip::
+
+  The above instructions work on *malformed* databases, but **if your SQLite3 version is 3.29 or newer**, you can use the `.recover command <https://sqlite.org/cli.html#recover>`__ to restore other types of database corruption as well. This can happen if Kolibri is terminated abruptly or if your storage media fails.
+
+  You can see your SQLite3 version by running ``sqlite3 --version`` from command line. If your database is corrupted but your SQLite3 is too old to run ``.recover``, consider copying your ``db.sqlite3`` file to a different system with a new SQLite3.
+
+  To apply ``.recover`` on a broken database, open up your Terminal and run a set of commands similar to when we applied ``.dump``:
```
radinamatic:
```suggestion
  To apply ``.recover`` on a broken database, open your Terminal and run the following commands (the sequence is similar to above where you used ``.dump``):
```

docs/manage/get_support.rst
```diff
@@ -56,16 +56,38 @@ The ``sqlite3`` command is necessary. This can be installed with ``sudo apt inst
...
+
+  .. code-block:: bash
```
radinamatic: Could you rework this code block to be separated commands instead (like I did with the first set of DB repair commands few weeks ago)? Remember that anything inside the `code-block` directive does not get localized, which leaves all the instructions in the comments not translated. 

I had the full intention to use the _Suggest_ feature, but it just becomes too messy to do in this editor, plus given that all is inside the `tip` block, I have no way of checking if the proper formatting is preserved  :confused:

radinamatic (changes requested):
It would be great to have the instructions localizable :wink:

docs/manage/get_support.rst
benjaoming: I found and tested a settting that marks both image alt texts and literal blocks for translation. This means that the code example now shows up for translation.

This adds a new challenge for literal blocks that previously worked in all languages: Getting translations to 100% means to generate zero-modification copies in Crowdin.

@radinamatic is it of any concern that these blocks are generally marked translatable? Won't translators for instance already know to not make mistakes in programmatic syntax and be sure to use copy-paste rather than typing manually?

image

radinamatic: Kudos to you for noticing that we weren't extracting the alternative text for images and translating them, that will be a great improvement and a welcome addition. :+1: :100: 

However, I'm not a fan at all of using this approach for literal code blocks. It makes the job unnecessarily more complicated for translators who have to **intermittently translate and leave unmodified the lines** in one same _presumably_ translatable segment...? 

And all for what exactly, just so we could save some text length by not repeating the ``.. code-block::`` directive? You yourself admitted that copy-pasting long lists of commands is a bit of a mal-practice. I would argue that clearly ordered steps with single line of commands in each have higher readability and promote better understanding of the code one is executing (and eventual errors that might be raised), compared to copy-pasting entire chunks.

benjaoming: Sounds good -- for this one-off case, I will adapt it to the chunked version to not mix commands into translatable contents.

It frees up screen space, makes it easier to write documentation and translate it as well. It also makes it better for the reader, as they get an option to copy-paste the entire instruction: Note that the above instruction actually was written for that exact purpose. That also makes it easier to test the instructions :)

Yes, it means that the translator has to know that they are translating syntax, and I think this might ultimately be a sort of soft requirement.. so either translators already know this, or they become familiar with it while translating. If not, we might implement automated steps or contextual instructions (you know those in-context messages for translators).

benjaoming:
Thanks for the review, @radinamatic -- I split up all of this in two sections now.

docs/manage/get_support.rst
```diff
@@ -56,16 +58,56 @@ The ``sqlite3`` command is necessary. This can be installed with ``sudo apt inst
...
+    .. code-block:: bash
+
+      mkdir -p malformed
+      cp -b db.sqlite3* malformed/
+
+#. Apply the ``.restore`` command to the database and move the restored database to the active location. Check the command line outputs for potential errors.
+   
+    .. code-block:: bash
+
+      sqlite3 db.sqlite3 .dump | sqlite3 fixed.db
```
radinamatic: Shouldn't it be ``.restore`` command here instead of ``.dump``...? :confused:

benjaoming: EEEEK

benjaoming: THANKS!

benjaoming:
@radinamatic do you have something available to verify the `.recover` step? Ubuntu 19.10 and up have SQLite 2.29+ https://packages.ubuntu.com/search?keywords=sqlite3

radinamatic:
Believe it or not, I never had that error appear to me locally (just once on a partner server I was upgrading remotely), how can I intentionally _malform_ a DB? :smiling_imp:

radinamatic (approved):
Looking good now, let's push it in! :+1:

benjaoming:
If the commands work on a non-corrupt database, that's also okay :)

PR #166, opened 2020-04-13
Add the updated strings for translation

radinamatic:
Generated the new strings with `gettext` (already uploaded to Crowdin), so the translation into Italian could begin.


benjaoming:
Pushing them to Crowd-in?

radinamatic:
Thank you, these strings are already on Crowdin! :slightly_smiling_face:

PR #174, opened 2020-05-04
Prefer `apt install` since 16.04

benjaoming:
Learned that `apt install` also works on .deb files since 16.04 and will install dependencies as well. Since we recommend `python3-cryptography`, this change could potentially ensure that people don't miss out on it.

~I need to make a test to see what happens when someone is offline and does not have `python3-cryptography`~ Tested: It will fail unless `--fix-missing` is added.

I added a tip box in the docs. I think this is the only place where we currently have information about why this package is a dependency. Installing Kolibri as a portable .deb can be a nice experience, just this little tip became more important since 20.04 runs Python 3.8 and won't benefit from bundled C extensions (cext).

image

### DIFF
--- a/docs/install/ubuntu-debian.rst
+++ b/docs/install/ubuntu-debian.rst
@@ -49,14 +49,14 @@ When you use the PPA installation method, upgrades to newer versions will be aut
 Install from a .deb file
 ------------------------
 
-The advantages of downloading a ``.deb`` file is the portability: you can copy the file from device to device and install Kolibri without internet access.
+The advantages of downloading a ``.deb`` file is the portability: You can copy the file from device to device and install Kolibri without internet access.
 
 #. Download the latest `.deb installer <https://learningequality.org/download/>`_ for Kolibri **version 0.13**, or have it copied to your local drive.
-#. Run this command from the location where you downloaded the ``DEB`` file:
+#. Run this command from the location where you downloaded the ``.deb`` file:
 
    .. code-block:: bash
 
-       sudo dpkg -i kolibri-installer-filename.deb
+       sudo apt install --fix-missing ./KOLIBRI_FILENAME.deb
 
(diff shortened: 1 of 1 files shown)


benjaoming:
This isn't optimal, but having `apt install --fix-missing ./kolibri.deb` command seems like the best choice: It will try to install `python3-cryptography` but gracefully timeout for DNS resolution of online repositories.

Using `dpkg -i kolibri.deb` will not fetch the dependency no matter what.

Important to note that `apt install ./kolibri.deb` will fail in an offline environment without either `python3-cryptography` already installed or `--fix-missing`:

VirtualBox_Ubuntu 20 04 server_05_05_2020_15_58_09

radinamatic:
If I understand correctly `apt install --fix-missing` is supposed to work offline the same as `dpkg -i`, correct? And `python3-cryptography` is not necessary?

benjaoming:
> And python3-cryptography is not necessary?

50% of the installer footprint size comes from bundling this dependency. I treat it as a decision that this is a highly desired dependency.

I think the "tip box" is now more important because 0.13 doesn't have cryptography in 20.04 and other Py 3.8 environments unless it's installed. So if someone is doing a portable installation, I think they should be informed to bring it with them... but okay it IS optional, so if this isn't clear enough, we shouldn't make people confused in a way that they give up. I hope the "tip" aspect makes it seem non-mandatory.

jredrejo:
> If I understand correctly `apt install --fix-missing` is supposed to work offline the same as `dpkg -i`, correct? 

No, `apt install -f` might require internet,  it's a way to fix a dependency problem. `dpkg -i `is a way to install a deb existing in your system, they have different purposes

benjaoming:
Yes, in addition to @jredrejo's correct point, `apt install --fix-missing` will gracefully handle network connection failure for soft dependencies.

PR #183, opened 2020-08-14
Integrated feedback from the Imps team, added new 'Facilities' icon

radinamatic:
self merge to check online before public release


docs/conf.py
```diff
@@ -227,6 +228,8 @@ def setup(app):
    :alt: Print
 .. |export| image:: /img/export.png
    :alt: Export
+.. |exams| image:: /img/quiz-icon.png
+   :alt: Quizzes
```
indirectlylit, line 232: see https://github.com/learningequality/kolibri-docs/pull/202 as an alternative to manually creating these

PR #202, opened 2020-09-15
updated icon handling in kolibri user docs

indirectlylit:
This PR is:
* supported by https://github.com/learningequality/kolibri-design-system/pull/104
* helping to address #183

It updates the way that the user docs displays icons:

Includes a new auto-generated file `rstIconReplacements.txt` that was generated from the design system and maps [design system icon aliases](https://kolibri-design-system.netlify.app/icons/) to design system SVGs and default colors

It also makes all icons `role="presentation"`, or semantically ignored by screenreaders. This is a good thing because previously, the text labels were not being internationalized.

| old | new | notes |
|--|--|--|
|  Click the → button to view more items | Click the → (next) button to view more items |  1 |
| Go to the **Reports** tab  (⬤ icon) to track learner progress | Go to the ⬤ **Reports** tab to track learner progress | 2 |

Notes:

1. Icons now have adjacent text in the docs to explicitly name the tab or button being referred to which will allow screen readers to work as expected and allow proper translation. It also helps sighted users associate a specific name with the icon
2. Conventionally always put the icon before the thing it describes when it appears this way in the UI, and uses that pattern consistently

Feedback is welcome!


radinamatic (approved):
Thank you for making the work around adding an updating the icons in the docs much easier! :100: 
Nice step towards Kolibri and its docs being more consistent :heart:

docs/coach/groups.rst
```diff
@@ -4,7 +4,7 @@
 Manage Groups
 -------------
 
-In case you need to further divide learners inside classes, for example to address the different progress needs, you can use the **Groups** feature. Create and delete groups, as well as assign learners to them from the **Groups** sub-tab in the **Plan** tab (|edit| icon) of your **Coach** dashboard. Default view displays the list of all groups for the selected class, with the number of assigned learners for each group.
+In case you need to further divide learners inside classes, for example to address the different progress needs, you can use the **Groups** feature. Create and delete groups, as well as assign learners to them from the **Groups** sub-tab in the |edit| **Plan** tab of your |coach| **Coach** dashboard. Default view displays the list of all groups for the selected class, with the number of assigned learners for each group.
```
indirectlylit, line 7: > `|edit| **Plan** tab`

FYI @radinamatic this is an example of what I was describing where the icon doesn't always map one-to-one with a label.

radinamatic: Yes, we'll have to do some clean-up and eventual renaming before trying to achieve 1:1 consistency :+1:

PR #208, opened 2021-02-08
 typofix in bulkimportusers CLI examples

ivanistheone:
small fix spotted via the twitters https://twitter.com/LearnEQ/status/1358802695537180676

relevant mgmt command source is https://github.com/learningequality/kolibri/blob/release-v0.14.x/kolibri/core/auth/management/commands/bulkimportusers.py#L354-L396 
which shows `filepath` is regular arg, not `--output-file` kwarg

### DIFF
--- a/docs/manage/command_line.rst
+++ b/docs/manage/command_line.rst
@@ -153,21 +153,21 @@ Execute the *dry-run* of the command to review the report containing the number 
 
 .. code-block:: bash
 
-  kolibri manage bulkimportusers --dryrun my-school-users-2020.csv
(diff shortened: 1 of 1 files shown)


indirectlylit:
hi @ivanistheone!

thanks for the edits. (as aside, we should have sent that user to the simpler GUI-based mechanism - https://kolibri.readthedocs.io/en/latest/manage/data.html#csv-import)

radinamatic (approved):
Thank you for catching that @ivanistheone!

We may rename those files in code snippets to avoid having to circle back next year :wink:

PR #223, opened 2021-05-27
learn-docs: change login to sign in

dufcrule:
Removed the term 'login' and replaced it with 'sign in' to maintain consistency. 'Sign in' is the term used across the rest of the Kolibri docs.


radinamatic (approved):
Good catch, thank you! 👍🏽

PR #230, opened 2021-09-26
Fix order of parameters in systemctl command

jredrejo:
https://kolibri.readthedocs.io/en/latest/install/raspberry_pi_manual.html has an error when using the systemctl command, order of parameters is incorrect.
This is causing problems with some users
https://community.learningequality.org/t/sudo-systemctl-kolibri-stop-returns-unknown-operation-kolibri/2181

This pr fixes it


indirectlylit (approved):
thank you!

PR #231, opened 2021-09-27
Add missing command

jredrejo:
The command name was missing in a couple of sentences.
`kolibri manage movedirectory needs` to be
`kolibri manage content movedirectory `

As reported by one of our forums user
https://community.learningequality.org/t/sudo-systemctl-kolibri-stop-returns-unknown-operation-kolibri/2181/2


radinamatic (approved):
Thanks for picking up on this! 🙏🏽

Issue #235, opened 2021-11-22
Remove manual instructions on how to setup kolibri in RPi


jredrejo:
Current manual instructions are quite problematic as they require some sysops skills.
But, after 0.15 is released, they're not valid any more. I think removing them is the best option because
- if the user has the needed sys skills he will know how to do it, specially if we give instructions on using kolibri-server
- If the user does not have the needed skills, changes in the kolibri options will probably break all the instructions and give us a bad image
- After 0.15 is released the current instructions are not valid anymore

jredrejo:
Closed in #236

PR #236, opened 2021-12-02
Fix requirements.txt & remove Raspberry manual setup

jredrejo:
Removing Raspberry Pi manual instructions because:
* Current Raspberry Pi manual setup is not valid anymore after v0.15 release
* Even when it was valid, it's a complex setup that requires important skills and it was a source of problems in the support forums

Also fixed Python requirements.txt to pin `sphinx_rtd_theme` because newer versions are incompatible with the current build setup

Closes: #235


docs/install/raspberry_pi.rst
```diff
@@ -37,4 +35,4 @@ By default the server does not have Internet access. To import resources to Koli
 Manual setup
 ============
 
-It is not necessary to use the custom OS image. If you already have a running Raspberry Pi set up and want to install Kolibri, or if you prefer to set up manually from an official Pi image, please refer to our :ref:`Raspberry Pi Tutorial for Kolibri <rpi_manual>`. 
+It is not necessary to use the custom OS image. If you already have a running Raspberry Pi set up and want to install Kolibri, or if you prefer to set up manually from an official Pi image, the installation of the ``kolibri-server`` package is recommended. The documentation of the :ref:`configuration is described here <kolibri_server_package>`. More complex configurations will require of Linux Sysop skills.
```
indirectlylit: I feel like this could use just a little elaboration, because as it's currently written `kolibri-server` sounds quicker and easier than using the image.

For example, it's unclear:

* What features does the custom image provide that are missing from `kolibri-server`?
* What features does `kolibri-server` provide that are missing from a fully-manual installation?
* What is meant by "complex configurations", and what specific Linux skills or knowledge areas would be necessary to replicate those?

jredrejo: mmm, @indirectlylit  I'll need your help to write a better explanation. 

- To install `kolibri-server` you have had to previously install and configure a raspberry image, usually from www.raspberry.org. So it's not easier as it requires several previous steps
- Our image provides an out-of-the-box full configurated experience, including not only kolibri-server but also a wifi hotspot and networking. No steps are needed once the SD card is inserted and the RPi is booted.

Using `kolibri-server` on an ALREADY configured RPi makes sense if you have previous things on your RPi disk and want to keep them, and it does not provide a hotspot. That's in the complex configuration areas. If the user does not want to use `kolibri-server` he'll also need to know about nginx, redis and uwsgi.

With all this bricks maybe you can help me to build a better hut

indirectlylit: Thank you for the bricks! I've assembled them here: https://github.com/jredrejo/kolibri-docs/pull/1

jredrejo: Thanks very much @indirectlylit , your patch is now in this PR

radinamatic (approved):
Bricks for the docs hut being laid down without me, loving it! 😍

PR #238, opened 2021-12-13
Docs updates for the 0.15 release

### LINKED ISSUE #68: Add Kolibri version as a substitution string and replace all hard-coded versions
There are many mentionings of versions. Both in links and instructions and file name examples. We can deal with these in various appropriate ways. Hard-coding isn't banned as such, just nice to avoid.

### LINKED ISSUE #219: Update rst icons once Learning Activities icons are in KDS
When https://github.com/learningequality/kolibri-design-system/issues/232 is completed, the `rstIconReplacements` file from KDS (https://github.com/learningequality/kolibri-design-system/blob/v0.2.x/docs/rstIconReplacements.txt) should be copy-pasted into the file of the same name in this repository.

### LINKED ISSUE #237: update macOS compatibility
mac OS compatibility is out-of-date, even for Kolibri 0.14:

image

I believe by now we've tested as high as macOS 12. We should also double-check that we're continuing to support those older versions and remove them if not.

radinamatic:
Summary of changes:

* added instructions and screenshots for the new features in 0.15
* updated the icons for all the new learning activities. Fixes #219.
* removed the hardcoded version mentions and replaced with substitution docstring. Fixes #68.
* added warning for permission problems during export (mentioned in https://github.com/learningequality/kolibri/issues/6685#issuecomment-603283760).
* left only Catalina (10.15) macOS compatibility, as that is the only version where we've been able to run the full critical workflows test plan. If we receive the confirmation that it runs without issues on more recent versions, we'll update. Fixes #237.
* changed all headings from **Title Case** to **Sentence case**.


docs/install/initial_setup.rst
```diff
@@ -75,7 +75,7 @@ Here are definitions of concepts necessary to understand the available options i
 
 		    * **Learn-only device**
 
-		      Unlike a 'full device', on a learn-only device only the features for learners are available. Coaches and admins can sign in but will only see the *Learn* page. Learn-only devices also include automated data syncing functionality.
+		      Unlike a 'full device', a learn-only device will have available only the features for learners. Coaches and admins can sign in but will only see the *Learn* page. Learn-only devices also include automated data syncing functionality.
```
indirectlylit: slight phrasing suggestion:

```suggestion
		      Unlike a 'full device', a 'learn-only' device will only have features for learners available. Coaches and admins can sign in but will only see the *Learn* page. Learn-only devices also include automated data syncing functionality.
```

radinamatic: I've been moving around the words in this sentence for a while, as that repetition: _...learn-**only** device will **only**..._ just screeches in my ears 😒

indirectlylit: haha!

PR #240, opened 2022-03-01
Add link to Running Kolibri on a different port when the default port is in use

SavvyOsive:
For someone initially setting up Kolibri, getting the default port 8080 already in use is a common issue, it will be better if a note is provided to the end users in advance with the reference to the doc for running Kolibri on a different port.

### DIFF
--- a/docs/access/linux.rst
+++ b/docs/access/linux.rst
@@ -18,7 +18,7 @@ Starting Kolibri on Linux will differ depending on the method you used to instal
       kolibri start
 
     .. note::
-      By default **Kolibri** starts on 8080 port. If the default port is in use, then refer to `this <https://kolibri.readthedocs.io/en/latest/manage/options_ini.html#run-kolibri-from-a-different-port>`_ doc for running it from a different port.
+      By default **Kolibri** starts on 8080 port. If the default port is in use, then refer to :ref:`this <different_port>` doc for running it from a different port.
 
   2. Open the default browser at ``http://127.0.0.1:8080``, and it will display the **Kolibri** start page.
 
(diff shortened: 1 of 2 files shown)


docs/access/linux.rst
```diff
@@ -16,7 +16,9 @@ Starting Kolibri on Linux will differ depending on the method you used to instal
     .. code-block:: bash
 
       kolibri start
-      
+
+    .. note::
+      By default **Kolibri** starts on 8080 port. If the default port is in use, then refer to `this <https://kolibri.readthedocs.io/en/latest/manage/options_ini.html#run-kolibri-from-a-different-port>`_ doc for running it from a different port.
```
indirectlylit: Just a heads-up that the link should use use an internal target – e.g. https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#the-ref-role – because otherwise the link will break when the path prefix and domain change.

indirectlylit (commented):
this is a great idea

docs/access/linux.rst
SavvyOsive: Yes, thanks for reviewing. I've updated by adding the cross reference to the doc. Please check now and let me know if any other change is required.

SavvyOsive: @indirectlylit please check

SavvyOsive:
Please check

SavvyOsive:
@indirectlylit please check

radinamatic:
@SavvyOsive Thank you for suggesting this improvement, and including the  change by @indirectlylit.
To make the link more accessible, please do not wrap one word only inside, but extend it a bit more:

```   
 .. note::
   By default **Kolibri** starts on 8080 port. If the default port is in use, refer to these :ref:`steps to change the port <different_port>`.
```

SavvyOsive:
@radinamatic @indirectlylit Please review now, thanks.

radinamatic (approved):
Looking good, thank you for contributing to Kolibri docs! 👍🏽

